PR #19, opened 2020-11-17
Feature/education and setup

matthewcarlreetz:
This PR completes the onboarding of a new user


package.json
```diff
@@ -54,17 +54,20 @@
...
     "@types/lodash": "^4.14.165",
+    "date-fns": "^2.16.1",
     "expo-haptics": "^8.3.0",
     "expo-secure-store": "^9.2.0",
     "i18next": "^19.8.3",
     "lodash": "^4.17.20",
     "lottie-ios": "3.1.8",
     "lottie-react-native": "^3.5.0",
     "react": "16.13.1",
+    "react-async-hook": "^3.6.2",
```
allenan, line 65: this is cool! hate doing async functions in `useEffect`

src/App.tsx
```diff
@@ -1,12 +1,17 @@
...
 import { SafeAreaProvider } from 'react-native-safe-area-context'
 import { ThemeProvider } from '@shopify/restyle'
 import { Platform, StatusBar } from 'react-native'
+import OneSignal from 'react-native-onesignal'
 import { theme } from './theme/theme'
 import NavigationRoot from './navigation/NavigationRoot'
 
 const App = () => {
+  useEffect(() => {
+    OneSignal.setAppId('64cb14bf-7121-4f41-997f-25e1917e6a1f')
```
allenan, line 12: these types of ids aren't intended to be secure, but is there any standard way to inject them as env vars at build time?

matthewcarlreetz: yes, thanks for reminding me. I pulled that over from the other app and forgot to come back to it.

src/navigation/NavigationRoot.tsx
```diff
@@ -20,22 +23,30 @@ const forFade = ({ current }: StackCardInterpolationProps) => ({
...
+  const dispatch = useAppDispatch()
 
   useEffect(() => {
+    dispatch(restoreUser())
     setTimeout(() => {
       setShowSplash(false)
-    }, 750)
-  }, [])
+    }, 1250)
+  }, [dispatch])
```
allenan, line 36: in what case does `dispatch` change value to trigger `useEffect` to execute again?

src/utils/account.ts
```diff
@@ -1,43 +1,64 @@
 import { Keypair, Mnemonic, Address } from '@helium/crypto-react-native'
 import * as SecureStore from 'expo-secure-store'
 
+type AccountStoreKey = BooleanKey | StringKey
+
+const stringKeys = ['mnemonic', 'keypair', 'address', 'userPin'] as const
+type StringKey = typeof stringKeys[number]
+
+const boolKeys = [
+  'accountBackedUp',
```
allenan, line 10: do we need these registered in `account.js` if they're also represented in redux?

src/utils/account.ts
```diff
@@ -1,43 +1,64 @@
...
+  'isEducated',
+  'isSettingUpHotspot',
+] as const
+type BooleanKey = typeof boolKeys[number]
+
+export const setItem = async (key: AccountStoreKey, val: string | boolean) => {
```
allenan, line 16: we may want to avoid exporting the arbitrary `setItem` function since that could encourage yolo setting/getting elsewhere in the app. it may be better to have dedicated functions within here

matthewcarlreetz: It's type checked, so you can't just go nuts. I like it this way because it avoids writing a ton of boilerplate functions. I could be convinced otherwise though.

Image 2020-11-17 at 4 22 06 PM

src/utils/account.ts
matthewcarlreetz: I like them being here even though it's a little redundant. The use case is when a user deletes and reinstalls the app. The keystore will be persisted whereas async-storage/redux-persist won't be.

package.json
matthewcarlreetz: same. It gets annoying and feels misplaced.

src/utils/account.ts
allenan: ok that works

src/utils/account.ts
allenan: that should be ok then, just want to make sure we account for the keys we're setting, but looks like this accomplishes that

src/navigation/NavigationRoot.tsx
matthewcarlreetz: It won't, but I'm just being "exhaustive" to appease eslint. I could just disable the eslint warning. It comes from the `eslint-plugin-react-hooks` library.

src/locales/en.ts
```diff
@@ -143,4 +143,260 @@ export default {
...
+      option_two: 'RAK Hotspot Miner',
+      third_party_header: 'Other Hotspots',
+      helium_edition: 'For the Helium Network',
+      fine_print:
+        'RAK Hotspot Miners have special firmware preloaded by RAK. Double check your hardware is a RAK Hotspot Miner before proceeding.',
+    },
+    start: {
+      title: 'Set up Hotspot',
+      subtitle_1:
+        'Helium Hotspot enables anyone to earn cryptocurrency, Helium (HLM), for providing wireless coverage for low power Internet of Things devices.',
```
allenan, line 159: this should be HNT, might be outdated in the current app too

allenan (approved):
🚀

Issue #15, opened 2020-11-12
Scaffold out 4 tab app structure


cokes518:
Start putting up the 4 tab structure:
Hotspot, Wallet, Notification, Settings

PR #27, opened 2020-11-20
Feature/tab nav

matthewcarlreetz:


src/navigation/mainTabs/MainTabNavigator.tsx
```diff
@@ -0,0 +1,46 @@
...
+              color={color}
+              size={size}
+            />
+          )
+        },
+      })}
+    >
+      <MainTab.Screen name="Account" component={Hotspots} />
+      <MainTab.Screen name="Hotspots" component={Hotspots} />
+      <MainTab.Screen name="Network" component={Hotspots} />
```
allenan, line 40: I know this is just a stub, but we probably want to steer in the direction of the new designs here. The four tabs in the new app are currently, Hotspots, Account, Notifications and More

image

allenan: I guess "Wallet" is what we'll probably call the second tab based on: https://github.com/helium/hotspot-app/issues/15

allenan (approved):
tabs look good, we'll update the icons in a future PR

Issue #32, opened 2020-11-24
Make Appv3 available on testflight


cokes518:
will need to be on a different release train than current app (v2).

cokes518:
Plan is to create a new app with new app ID, get a TestFlight going, then when we're ready to go to prod, update the app ID.

cokes518:
This is done.

PR #35, opened 2020-11-30
new identifier for testing

allenan:
Changes app id as mentioned in https://github.com/helium/hotspot-app/issues/32 so that we can set up a release train on appcenter that won't overwrite the current app, for now.


matthewcarlreetz (approved):
Looks good to me.

PR #38, opened 2020-12-04
Feature/hotspot onboarding

matthewcarlreetz:
This pr onboards a hotspot to the point of connecting and configuring it.


src/providers/HotspotProvider.tsx
```diff
@@ -0,0 +1,26 @@
...
+  connectedHotspot: undefined,
+  availableHotspots: {},
+}
+
+const HotspotContext = createContext<ReturnType<typeof useHotspot>>(
+  initialState,
+)
+const { Provider } = HotspotContext
+
+const HotspotProvider = ({ children }: { children: ReactNode }) => {
```
allenan, line 20: it might be a little unclear just from a naming perspective that this is providing a context for the hotspot that you're currently connected to over bluetooth

matthewcarlreetz: I can definitely see that. Any suggestions? `HotspotBleProvider`, `HotspotDeviceProvider`, ? @allenan

src/features/hotspots/setup/HotspotSelectionScreen.tsx
```diff
@@ -0,0 +1,105 @@
...
+        numberOfLines={2}
+        adjustsFontSizeToFit
+        marginVertical="l"
+      >
+        {t('hotspot_setup.selection.subtitle')}
+      </Text>
+
+      <Box flexDirection="row" height={191}>
+        <Card flex={1} variant="elevated" backgroundColor="white">
+          <TouchableHighlightBox
```
allenan, line 33: I think the icon should be tinted white during the active state here

src/providers/HotspotProvider.tsx
allenan: maybe `ConnectedHotspotProvider`?

matthewcarlreetz: Cool. I'll go with that.

src/features/hotspots/setup/HotspotSelectionScreen.tsx
matthewcarlreetz: Ok, I'll do that as well as remove react-native-vector-icons as part of this PR.

allenan (approved):
looks good to me! I'm sure the bluetooth stuff will evolve as we get more of that flow ported through, but overall happy with the provider/hook setup

PR #53, opened 2020-12-16
Functionally finish remaining onboarding screens

matthewcarlreetz:


src/features/hotspots/setup/ConfirmLocationScreen.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  const onMapMoved = async (newCoords: number[]) => {
+    setMarkerCenter(newCoords)
+    setMapMoving(false)
+
+    const [longitude, latitude] = newCoords
+    const [{ street, city }] = await reverseGeocode(latitude, longitude)
+    const name = street && city ? [street, city].join(', ') : 'Loading...'
+    setLocationName(name)
+
```
matthewcarlreetz, line 31: @tyler-whitman @allenan ↓↓

src/features/hotspots/setup/HotspotSetupAddTxnScreen.tsx
```diff
@@ -0,0 +1,168 @@
...
+      const txn = await addGatewayTxn()
+      setSubmitting(false)
+      if (txn) {
+        navNext()
+      }
+    } else {
+      navNext()
+    }
+  }
+
```
matthewcarlreetz, line 55: @tyler-whitman @allenan ↓↓ Can you check the logic here?

src/utils/bluetooth/bluetoothDataParser.ts
```diff
@@ -0,0 +1,80 @@
+import { decode } from 'base-64'
+import {
+  wifi_services_v1 as WifiServicesV1,
+  wifi_remove_v1 as WifiRemoveV1,
+  wifi_connect_v1 as WifiConnectV1,
+  add_gateway_v1 as AddGatewayV1,
+} from '@helium/proto-ble'
+import { util } from 'protobufjs'
+import { HotspotCharacteristic, FirmwareCharacteristic } from './bluetoothTypes'
+
```
matthewcarlreetz, line 10: @tyler-whitman @allenan ↓↓ Do you know what this array check is here for? Is it needed still?

src/features/hotspots/setup/ConfirmLocationScreen.tsx
allenan: I think that was to make the next button be disabled for 3 seconds or until the user moves the map. We had a problem where users were just pressing "next" "next" "next" to get through onboarding and ended up asserting in the wrong location

matthewcarlreetz: oh, ok, that makes sense. I'll add that in.

src/features/hotspots/setup/HotspotSetupAddTxnScreen.tsx
```diff
@@ -0,0 +1,168 @@
...
+    // TODO: Verify this is correct
+    if (freeAddHotspot) {
+      setFee(0)
+      return
+    }
+
+    const calc = async () => {
+      const nonce = account?.nonce ? account.nonce + 1 : 0
+      // TODO:                                            ↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓
+      const val = await calculatePaymentTxnFee(0, nonce, 'what should this be?')
```
allenan, line 66: I think we'd want to be calculating the fee for an `add_gateway_v1` txn rather than payment. But since txn fee is based on byte size of the serialized txn, I think I just hardcoded the fee and staking amount for add and assert, because it'll always be the same. Payment txn fee can change depending on how large the `amount` integer is and how many recipients you add (although our app only supports one recipient right now)

src/utils/bluetooth/bluetoothDataParser.ts
allenan: I think that was to check that the response from the hotspot was well-formed, and not an error or some garbage string

src/features/hotspots/setup/HotspotSetupAddTxnScreen.tsx
allenan: I think for now our logic can be if it's "free" (ie it came back as registered in the staking server, so helium, inc pays the dc, making it free from the user's perspective), then show the 0 fee screen. If not, then show the hardcoded fee screen. We may want to even change what that fee screen shows in that case because even if you had enough DC to cover that yourself, you wouldn't be able to add it because it's not made by a helium-approved manufacturer. So as long as we just get the free vs not-free logic working for now, that's probably fine. We can revise a bit later

src/features/hotspots/setup/HotspotSetupAddTxnScreen.tsx
matthewcarlreetz: I just looked back at the old app. The fee calculation looks like this
```
  calculateFee = () => {
    const {
      data: {
        account: { fee },
      },
      bluetooth: { connectedHotspotDetails },
    } = this.props
    if (connectedHotspotDetails.free_add_hotspot) return 0
    return fee
  }
  ```
  
But, [fee in the account data is hardcoded to 0](https://github.com/helium/mobile-wallet/blob/master/src/data/AppData.js#L311)

Can I assume free at this point?

Comment (allenan) [reply src/features/hotspots/setup/HotspotSetupAddTxnScreen.tsx:66]: yeah I mean to the user, the `add_gateway` fee will always be 0, for now. Either they're allowed to proceed or not

File: package.json [hunk 5]
```diff
@@ -83,6 +89,7 @@
     "lodash": "^4.17.20",
     "lottie-ios": "3.1.8",
     "lottie-react-native": "^3.5.0",
+    "protobufjs": "^6.10.2",
```
tyler-whitman, line 92: Did we still need this one even with `@helium/proto-ble`?

src/utils/appDataClient.ts
```diff
@@ -3,12 +3,44 @@ import { getString } from './account'
...
   return newHotspotList.takeJSON(1000)
 }
 
+export const getHotspotDetails = async (address: string) => {
+  const { data } = await client.hotspots.get(address)
+  return data
+}
+
+export const getAccount = async () => {
+  const address = await getString('address')
```
allenan, line 20: I just realized `getString` here reads a bit odd. I know it's not explicitly part of this PR but it took me a minute to figure out what it meant in this context. maybe something like `getSecureValue(key: string)`?

src/utils/crypto.ts
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable import/prefer-default-export */
+import Sodium from 'react-native-sodium'
+import { getString } from './account'
+
+export const sign = async (message: string) => {
```
allenan, line 5: do we still need to do any signing here in the app? Ideally all of this happens through helium-js

src/utils/currency.ts
```diff
@@ -0,0 +1,35 @@
...
+export const abbrFormat = (num: number) => {
+  let formatted
+  if (num % 1 !== 0) formatted = `${round(num, 2)}`
+  if (num > 999) formatted = `${round(num / 1000, 1)}k`
+  if (num > 999999) formatted = `${round(num / 1000000, 1)}m`
+  if (num > 999999999) formatted = `${round(num / 1000000000, 1)}b`
+  return formatted
+}
+
+export const currencyFormat = (bones: number, maxDecimalPlaces?: number) => {
```
allenan, line 29: I think all currency formatting should be handled through the `Balance#toString` method in helium-js. If any fields are missing the conversion to `Balance` objects we should fix that

allenan (changes requested):
This looks really good! mainly just want to make sure that we're not duplicating anything that helium-js can do now that we didn't have before in the app, currency formatting and signing txns being the main ones

package.json
matthewcarlreetz: yeah, I use one of their utilities to do some decoding in `src/utils/bluetooth/bluetoothDataParser.ts`

src/utils/appDataClient.ts
matthewcarlreetz: yeah, I am totally with you on that.

src/utils/crypto.ts
matthewcarlreetz: This bit was copied over from the previous app. Are you thinking I can just do
```
export const makeAddGatewayTxn = async (partialTxnBin: string) => {
  const addGatewayTxn = AddGatewayV1.fromString(partialTxnBin)
  const serialized = addGatewayTxn.serialize()
  return Buffer.from(serialized).toString('base64')
}
```
in `src/utils/transactions.ts`?

allenan: you should be able to use the `sign` function here: https://github.com/helium/helium-js/blob/master/packages/transactions/src/AddGatewayV1.ts#L118

src/features/hotspots/setup/HotspotSetupAddTxnScreen.tsx
```diff
@@ -0,0 +1,141 @@
...
+import BackScreen from '../../../components/BackScreen'
+import Box from '../../../components/Box'
+import Button from '../../../components/Button'
+import Text from '../../../components/Text'
+import { useConnectedHotspotContext } from '../../../providers/ConnectedHotspotProvider'
+import { RootState } from '../../../store/rootReducer'
+import usePermissionManager from '../../../utils/usePermissionManager'
+import { HotspotSetupNavigationProp } from './hotspotSetupTypes'
+
+const FEE = 40
```
matthewcarlreetz, line 14: What this fee should be is the biggest thing I'm still not clear on 🤦 
It's currently set to `0` if `freeAddHotspot` otherwise `40`.

allenan: yeah I think I'll need to take a pass through all of that at some point, so as long as the general flow is working for now, let's not worry too much about the exact values. the blockchain won't accept anything that's invalidly constructed anyway

matthewcarlreetz: Sounds good. 👍

allenan (approved):
send it

PR #57, opened 2020-12-17
Feature/notifications

matthewcarlreetz:


src/features/notifications/NoNotifications.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+import Text from '../../components/Text'
+import BellBubble from '../../assets/images/bellBubble.svg'
+import Box from '../../components/Box'
+
+const NoNotifications = () => {
```
allenan, line 7: could maybe call this `EmptyNotifications` or something? The `NoNo` in `NoNotifications` kind of runs together when I skim it 😅

src/utils/secureAccount.ts
```diff
@@ -82,5 +83,46 @@ export const getKeypair = async (): Promise<Keypair | undefined> => {
   }
 }
 
+const sign = async (message: string) => {
+  // TODO: is there a better way to sign?
```
allenan, line 87: can we use the `Keypair#sign` function?

src/features/notifications/NoNotifications.tsx
matthewcarlreetz: ok, I'll rename it in my next PR, if that's all right.

src/store/account/accountSlice.ts
```diff
@@ -6,30 +6,58 @@ import {
...
   'account/fetchData',
   async () => {
-    try {
-      const data = await Promise.all([getHotspots(), getAccount()])
-      return { hotspots: data[0] || [], account: data[1] }
-    } catch (e) {
-      console.log(e)
-      return { hotspots: [] }
+    const data = await Promise.all(
+      [getHotspots(), getAccount(), getWallet('notifications')].map((p) =>
```
tyler-whitman, line 51: So does this just load once on app start?

src/features/notifications/NoNotifications.tsx
allenan: np

src/store/account/accountSlice.ts
matthewcarlreetz: for the moment, yes, but we will need to figure out a strategy around refreshing it.

tyler-whitman (approved):

PR #58, opened 2020-12-18
Show notifications list/view

matthewcarlreetz:
Show notifications in a grouped list.
Make notification items touchable
Create modal notification show container.
Finish notification show layout.
Create useVisible hook to track view visibility
Mark notifications as read when tab disappears
Refresh notifications when tab becomes visible


src/features/notifications/NotificationShow.tsx
```diff
@@ -0,0 +1,104 @@
...
+        <Card
+          variant="modal"
+          padding="l"
+          margin="m"
+          style={{ marginTop: insets.top + spacing.xxl }}
+        >
+          <Box flexDirection="row" alignItems="center">
+            <HeliumNotification />
+            <Text variant="body2" color="purpleMain" marginLeft="xs" flex={1}>
+              Helium Update
```
tyler-whitman, line 61: Move this one to translations?

src/features/notifications/NotificationList.tsx
```diff
@@ -1,13 +1,83 @@
...
+        keyExtractor={(item, idx) => `${item[0].id}.${idx}`}
+        renderItem={({ item, index }) => (
+          <NotificationGroup
+            index={Math.min(groupedNotifications.length - 1, 10) - index}
+            notifications={item}
+            onNotificationSelected={setSelectedNotification}
+          />
+        )}
+      />
+      <NotificationShow
```
allenan, line 76: I think in a future PR we could make sure of holy-sheet for these as well so that you'd get them animated up from the bottom, a scrollview interior, and swipe down to close.

src/features/notifications/NotificationsScreen.tsx
```diff
@@ -1,23 +1,48 @@
...
 
 const NotificationsScreen = () => {
   const {
-    account: { notifications },
+    account: { notifications, markNotificationStatus },
   } = useSelector((state: RootState) => state)
+  const dispatch = useAppDispatch()
+
+  const markAsRead = useCallback(() => {
+    if (!notifications.find((n) => !n.viewed_at)) return
```
allenan, line 19: I noticed that I had to reload the app to get the `new` indicator to disappear from my notifications

src/features/notifications/NotificationItem.tsx
```diff
@@ -0,0 +1,82 @@
...
+  isLast: boolean
+  onNotificationSelected: (notification: Notification) => void
+}
+const NotificationItem = ({
+  notification,
+  isFirst,
+  isLast,
+  onNotificationSelected,
+}: Props) => {
+  const isNew = !notification.viewed_at
```
allenan, line 21: should we also flag off the green circle in the bottom tab svg unless you have unread notifications?

src/features/notifications/NotificationList.tsx
matthewcarlreetz: That was my initial thought, but the design is a little different here than the rest of the app. There's a margin around the card and a blurred background. Whatever the case, I think I need to have a conversation with Pete around his vision for all of this.

src/features/notifications/NotificationsScreen.tsx
matthewcarlreetz: Interesting. I'll try to reproduce.

src/features/notifications/NotificationItem.tsx
matthewcarlreetz: oh, yeah, good call.

src/features/notifications/NotificationList.tsx
allenan: yeah I was thinking about adding a blurred background option directly in that library. and I think you should be able to override the styles around the container to give it some margin

src/features/notifications/NotificationsScreen.tsx
allenan: if it's not reproducible it could have just been something weird with my dev env

src/features/notifications/NotificationList.tsx
allenan: happy to take a pass on that at some point in the future

allenan (approved):
looking good!

PR #64, opened 2020-12-30
Feature/onboarding style updates

### LINKED ISSUE #31: New designs for account onboarding flow
Same flow as existing app, but new look and feel to match new designs.

matthewcarlreetz:
Closes #31


android/app/build.gradle
```diff
@@ -90,7 +90,7 @@ import com.android.build.OutputFile
  */
 
 project.ext.react = [
-    enableHermes: false,  // clean and rebuild if changing
+    enableHermes:true,  // clean and rebuild if changing
```
tyler-whitman, line 93: Nice I think this is supposed to be better for Android?

src/components/BackScreen.tsx
```diff
@@ -22,7 +22,7 @@ const BackScreen = ({
       backgroundColor={backgroundColor || 'primaryBackground'}
       flex={1}
     >
-      <BackButton paddingHorizontal="lx" onPress={navigation.goBack} />
+      <BackButton marginHorizontal="n_s" onPress={navigation.goBack} />
```
tyler-whitman, line 25: what is `n_s` and `lx` ?

tyler-whitman:
Screen Shot 2020-12-30 at 5 31 36 PM
Something is off with the colors in the account recovery

android/app/build.gradle
matthewcarlreetz: Yeah, it's a JS engine optimized for Android. It is required in order to use Reanimated2. We'll see...

src/components/BackScreen.tsx
matthewcarlreetz: n stands for negative, so negative_small. lx is in between large and extra large.

```
  spacing: {
    n_xxl: -60,
    n_xl: -40,
    n_lx: -32,
    n_l: -24,
    n_lm: -20,
    n_m: -16,
    n_ms: -12,
    n_s: -8,
    n_xs: -4,
    n_xxs: -2,
    n_xxxs: -1,
    none: 0,
    xxxs: 1,
    xxs: 2,
    xs: 4,
    s: 8,
    ms: 12,
    m: 16,
    lm: 20,
    l: 24,
    lx: 32,
    xl: 40,
    xxl: 60,
  }
```

matthewcarlreetz:
> Screen Shot 2020-12-30 at 5 31 36 PM
> Something is off with the colors in the account recovery

Good catch. I think we're doing account recovery as a separate PR because there aren't any updated designs for it right now. I'll fix the colors as part of this PR though.

allenan:
> > Screen Shot 2020-12-30 at 5 31 36 PM
> > Something is off with the colors in the account recovery
> 
> Good catch. I think we're doing account recovery as a separate PR because there aren't any updated designs for it right now. I'll fix the colors as part of this PR though.

Is it also expected right now that it's not autocompleting the possible words? It's interesting because I see it narrow down the potential number, but they all say whatever is in the input box, like "Te" above

matthewcarlreetz:
> > > Screen Shot 2020-12-30 at 5 31 36 PM
> > > Something is off with the colors in the account recovery
> > 
> > 
> > Good catch. I think we're doing account recovery as a separate PR because there aren't any updated designs for it right now. I'll fix the colors as part of this PR though.
> 
> Is it also expected right now that it's not autocompleting the possible words? It's interesting because I see it narrow down the potential number, but they all say whatever is in the input box, like "Te" above

Hmmm. I thought I fixed this. Do you have [this](https://github.com/helium/hotspot-app/pull/64/commits/172f534f48c14d99c4721b550dbe5ebdcb22877f#diff-35bc71bed6fdd71b38c32003379c1457be836b484b383c9f85f67864513d4ac9) in your local files? @allenan

allenan:
on the word carousel, the "I have written these down" never enables for me, even when I'm on the final word

allenan:
> on the word carousel, the "I have written these down" never enables for me, even when I'm on the final word

I see, I had to let it sit on each one, which is maybe fine

matthewcarlreetz:
> on the word carousel, the "I have written these down" never enables for me, even when I'm on the final word

It's weird... You have to let it actually stop on each word, so that `onSnapToItem` gets triggered. It works a lot better on an actual device than the simulator. If it's too finicky, we can just check to see if they've scrolled to the last word.

allenan:
> > on the word carousel, the "I have written these down" never enables for me, even when I'm on the final word
> 
> It's weird... You have to let it actually stop on each word, so that `onSnapToItem` gets triggered. It works a lot better on an actual device than the simulator. If it's too finicky, we can just check to see if they've scrolled to the last word.

I did it on an actual device, but I was just trying to get past it :)

If I were actually writing them all down, it should work fine

allenan:
image

noticed some off-center spacing here ^

matthewcarlreetz:
> image
> 
> noticed some off-center spacing here ^

Does that look better?
Simulator Screen Shot - iPhone 11 - 2021-01-04 at 13 50 39

allenan:
yeah that looks better to me

matthewcarlreetz:
> yeah that looks better to me

cool, I pushed the change.

allenan:
> > yeah that looks better to me
> 
> cool, I pushed the change.

hm not seeing a new commit

matthewcarlreetz:
> > > yeah that looks better to me
> > 
> > 
> > cool, I pushed the change.
> 
> hm not seeing a new commit

sorry. It's there now.

tyler-whitman (approved):
Looks good!

allenan (approved):
this looks good to merge

Issue #22, opened 2020-11-18
Payment Transactions


cokes518:
Porting over functionality from the old app:
**Receive**

- [x] generate and display public address in QR code format
- [x] ability to copy public address to clipboard
- [x] share card

**Send**

- [x] scan QR code to populate fields
- [x] show balance available
- [x] verify helium address
- [x] show fee
- [x] use number pad to enter amount
- [x] verify amount + fee does not exceed balance
- [x] implement Send Max feature
- [x] construct, sign, and submit transaction to blockchain
- [x] use Lottie animation

PR #77, opened 2021-01-06
Send Screen: handle scanning dc burn and payment QR codes

allenan:
In an attempt to keep these PRs smaller, this addresses handling both dc burn and payment fields, and scanning QR codes to populate those fields.

for example, here's a dc burn qr code:
image

and a payment that requests a specific amount:
image

and simply an address:
image


allenan:
this addresses parts of https://github.com/helium/hotspot-app/issues/22 but does not close it yet

src/features/wallet/send/SendLockedField.tsx
```diff
@@ -0,0 +1,34 @@
...
+import Text from '../../../components/Text'
+import InputLock from '../../../assets/images/input-lock.svg'
+
+type Props = {
+  label: string
+  value: string
+  bottom?: boolean
+}
+
+const SendLockedField = ({ label, value, bottom = false }: Props) => {
```
tyler-whitman, line 12: This seems generic enough to make it `LockedInput` or something like that. We will need it for transfer too

src/features/wallet/send/SendForm.tsx
```diff
@@ -0,0 +1,187 @@
...
+            </Text>
+          </TouchableOpacityBox>
+        }
+      />
+    </Box>
+  )
+
+  const renderBurnForm = () => (
+    <Box>
+      <SendInput
```
tyler-whitman, line 117: This can probably be generic like `LableInput` too?

src/features/wallet/send/SendHeader.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const containerStyles = useAnimatedStyle(() => {
+    return {
+      height: withTiming(
+        keyboardState.value === 1 ? 100 : 250,
+        animatedOptions,
+      ),
+    }
+  })
+
+  const miniHeaderStyles = useAnimatedStyle(() => {
```
tyler-whitman, line 61: Nice, some real animations for when the keyboard is up

src/features/wallet/send/SendLockedHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+import Box from '../../../components/Box'
+import Close from '../../../assets/images/close.svg'
+import TouchableOpacityBox from '../../../components/TouchableOpacityBox'
+import Text from '../../../components/Text'
+
+type Props = {
+  onClosePress: () => void
+}
+
+const SendLockedHeader = ({ onClosePress }: Props) => {
```
tyler-whitman, line 12: Could probably remove Send from this one too if we go with a generic approach

allenan: sure I can make those more generic

tyler-whitman (approved):
Looks good

PR #78, opened 2021-01-06
Feature/wallet data

matthewcarlreetz:


src/features/wallet/root/ActivityItem.tsx
```diff
@@ -1,49 +1,175 @@
...
   isLast = false,
+  address = '',
 }: Props) => {
-  const handlePress = () => {
-    triggerNavHaptic()
-  }
+  const colors = useColors()
+  const { t } = useTranslation()
+
+  const isSending = useMemo(() => isPayer(address, item), [address, item])
```
tyler-whitman, line 54: Something is up with this. Mine say Received but they are actually Sent payments

tyler-whitman: Actually it looks like its always false

src/features/wallet/root/ActivityItem.tsx
```diff
@@ -1,49 +1,175 @@
...
+
+  const icon = useCallback(
+    (type: string) => {
+      if (!TxnTypeKeys.find((k) => k === type)) {
+        return null
+      }
+
+      const size = 24
+      switch (type as TxnType) {
+        case 'transfer_hotspot_v1':
```
tyler-whitman, line 152: Should we save these types in an enum or something somewhere? Or do they type match already?

matthewcarlreetz: TxnType gives us type safety here.

src/features/wallet/root/ActivityItem.tsx
matthewcarlreetz: ok, I'm going to try some other addresses and see if I can replicate.

src/App.tsx
```diff
@@ -41,30 +42,37 @@ const App = () => {
       isRestored,
       isBackedUp,
       isRequestingPermission,
+      isLocked,
+      appStateStatus,
     },
   } = useSelector((state: RootState) => state)
 
+  useEffect(() => {
```
tyler-whitman, line 50: Does this fix something? Just wondering. I see we moved from using the direct callback to using the state and effect instead. Makes sense just wondering if it caused a bug or something?

src/App.tsx
```diff
@@ -73,6 +81,12 @@ const App = () => {
     dispatch(fetchData())
   }, [isRestored, isBackedUp, dispatch])
 
+  useEffect(() => {
+    if (isBackedUp) {
```
tyler-whitman, line 85: Whats the flow on this? It seems like something that should happen once on launch of the app? 

Why do we do it only if `isBackedUp` is true? And we need to run again if it goes false and back to true. Im wondering if just passing [] would work?

src/App.tsx
matthewcarlreetz: There was a bug on android where face id would send the app into 'background' which was causing an infinite loop of lock=>auth=>lock=>auth

src/App.tsx
matthewcarlreetz: We need to wait for it to be backed up so that the account address is defined. `client.account(address).activity.list()` requires account, so we reinitialize them whenever `isBacked` changes to true.

src/features/wallet/root/ActivityCard.tsx
```diff
@@ -34,51 +88,75 @@ const ActivityCard = forwardRef(
...
+            if (isPendingTransaction(item)) {
+              return `${filter}.${(item as PendingTransaction).hash}`
+            }
+
+            return `${filter}.${(item as AddGatewayV1).hash}`
+          },
           renderItem,
+          refreshControl: (
+            <RefreshControl
+              refreshing={txnStatus === 'pending' && !transactionData.length}
```
tyler-whitman, line 150: I dont see this when switching filters as it loads. I like the animation but I think it needs some kind of loading indicator or it feels like its frozen for a bit when loading transactions after changing filters.

Maybe we can add another boolean when the filter changes and its loading? We could also do this in a separate PR if you want as it works well already.

src/features/wallet/root/ActivityCardHeader.tsx
```diff
@@ -1,32 +1,63 @@
...
 
-const ActivityCardHeader = () => {
-  const handlePress = () => {
-    triggerNavHaptic()
+type Props = {
+  filter: FilterType
+  onFilterChanged: (filter: FilterType) => void
+}
+
+const ActivityCardHeader = ({ filter, onFilterChanged }: Props) => {
```
tyler-whitman, line 17: Not sure if it would be a lot of work, but we will need this in other places too so making it generic would be awesome.

Maybe instead of using specific filters we could make it a generic string array? We could also change it if we want when we need to use it in another place?

src/features/wallet/root/ActivityItem.tsx
tyler-whitman: Its fixed now

src/features/wallet/root/ActivityItem.tsx
tyler-whitman: Cool, works for me

src/features/wallet/root/ActivityItem.tsx
```diff
@@ -63,34 +196,38 @@ const ActivityItem = ({
...
-          </Text>
+          {time && <Text color="graySteel">{time}</Text>}
         </Box>
       </Box>
     </TouchableWithoutFeedback>
   )
 }
 
-export default ActivityItem
+export default memo(ActivityItem)
```
tyler-whitman, line 233: Nice, I think we should be using this for most future components?

src/App.tsx
tyler-whitman: Ah nice good find

src/App.tsx
tyler-whitman: Ah ok, that makes sense

tyler-whitman (approved):

src/features/wallet/root/ActivityItem.tsx
matthewcarlreetz: definitely for flatlist items

src/features/wallet/root/ActivityCard.tsx
matthewcarlreetz: hmmm. Are you testing on ios or android?

src/features/wallet/root/ActivityCardHeader.tsx
matthewcarlreetz: yeah, I'm with you on that. For now, I'd rather leave it as a todo

src/features/wallet/root/ActivityCard.tsx
tyler-whitman: The iOS simulator

src/features/wallet/root/ActivityCardHeader.tsx
tyler-whitman: Yeah sounds good

PR #79, opened 2021-01-07
owned hotspots map

tyler-whitman:
Adds owned hotspot markers to the map and centers to their location on scroll of the owned hotspot list.

This also fixes the lag caused by the map.


allenan (approved):
looks good!

matthewcarlreetz (approved):
LGTM!

PR #86, opened 2021-01-08
Finish send screen

### LINKED ISSUE #22: Payment Transactions

allenan:
This closes #22


src/features/wallet/root/BalanceCard.tsx
```diff
@@ -47,7 +57,7 @@ const BalanceCard = ({ onReceivePress, onSendPress }: Props) => {
               opacity={0.4}
               lineHeight={25}
             >
-              .45876891 HNT
+              .{decimalPart}
             </Text>
```
tyler-whitman, line 61: Should we update the account DC and HST balance?

src/features/wallet/send/SendView.tsx
```diff
@@ -43,39 +123,61 @@ const SendView = ({ scanResult }: { scanResult?: QrScanResult }) => {
...
+    if (type === 'dc_burn') {
+      return makeBurnTxn(getIntegerAmount(), address, getNonce(), memo)
+    }
+
+    throw new Error('Unsupported transaction type')
+  }
+
+  const handleSubmit = async () => {
+    const txn = await constructTxn()
+    await submitTransaction(txn)
```
tyler-whitman, line 155: Do we need to handle any errors here?

tyler-whitman (approved):
Looking good! 

We will need to handle pending transactions and errors. Also the balance doesn't update automatically once the transactions clears.

Do we have an idea for failed transactions yet? We definitely need to handle those better for when transactions don't clear.

src/features/wallet/root/BalanceCard.tsx
allenan: I was going to do that in a separate PR because it'll affect the animation points of the bottom sheet, but yes we should

allenan:
> Looking good!
> 
> We will need to handle pending transactions and errors. Also the balance doesn't update automatically once the transactions clears.
> 
> Do we have an idea for failed transactions yet? We definitely need to handle those better for when transactions don't clear.

Definitely. I think all of those points should be covered by the following issues:

- #94 
- #99 
- #100

src/features/wallet/send/SendView.tsx
allenan: do you think we should just do a `Logger.error` for that?

src/features/wallet/root/BalanceCard.tsx
tyler-whitman: Ok no problem, sounds good

src/features/wallet/send/SendView.tsx
tyler-whitman: I think a Logger error would be good, and also maybe show a dialog that something went wrong at least? Or the app will just be stuck right?

I guess its pretty rare that the construct and submit itself would fail though?

allenan: I can add a logger and an alert

PR #96, opened 2021-01-14
Feature/diagnostics

matthewcarlreetz:
https://user-images.githubusercontent.com/1760821/104651864-dd61b180-567d-11eb-9531-62757d8a2c16.mov

IMG_DF8086CEACD8-1


ios/Podfile.lock
```diff
@@ -791,4 +791,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: cb0b1503fafdfd47c00196486ddb9af228e8c1da
 
-COCOAPODS: 1.9.3
+COCOAPODS: 1.10.1
```
tyler-whitman, line 794: I think we need to keep this 1.9.3 for app center?

src/features/hotspots/settings/HotspotDiagnosticsConnection.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const hotspotCount = Object.keys(availableHotspots).length
+  const keys = Object.keys(availableHotspots)
+  const { t } = useTranslation()
+
+  const { connectedHotspot } = useSelector((state: RootState) => state)
+
+  useEffect(() => {
+    const handleConnect = async () => {
+      if (selectedHotspot && !!connectedHotspot.address) {
+        await sleep(500)
```
tyler-whitman, line 43: Would there be any difference here is doing a setTimeout() instead of a sleep? Did this come from the old code?

src/features/hotspots/settings/HotspotItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+      disabled={connected || connecting}
+    >
+      <HotspotIco color={selected ? white : purpleMain} />
+      <Text
+        variant="body1"
+        color={selected ? 'white' : 'black'}
+        marginLeft="ms"
+        flex={1}
+      >
+        {hotspot.localName || hotspot.name}
```
tyler-whitman, line 59: We were parsing out "Helium" to avoid confusion for RAK hotspots and any future ones that may come up. So it would be just "Hotspot 012A" or whatever. @allenan do you think we still want to do that?

tyler-whitman: Was just doing that btw `export const formatBluetoothName = (name) => name.replace('Helium', '').trim()`

tyler-whitman (approved):
Looks great

src/features/hotspots/settings/HotspotItem.tsx
cokes518: It should be "Hotspot 1234" if possible.

src/features/hotspots/settings/HotspotDiagnosticsConnection.tsx
matthewcarlreetz: I think it would have the same effect. `sleep` is a function I added. It's just a wrapper around setTimeout.

`export default (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))`

src/features/hotspots/settings/HotspotItem.tsx
matthewcarlreetz: ok, will do!

ios/Podfile.lock
matthewcarlreetz: Good catch. I don't know why my local cocoa pods install keeps auto-updating.

src/features/hotspots/settings/HotspotDiagnosticsConnection.tsx
tyler-whitman: Ah cool, looks good.

PR #101, opened 2021-01-19
Fix Android Release Build

tyler-whitman:
remove deprecated @react-native-community/picker

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -375,7 +375,7 @@ PODS:
     - React-Core
   - RNCMaskedView (0.1.10):
     - React
-  - RNCPicker (1.8.1):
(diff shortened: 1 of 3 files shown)


allenan:
should we remove the other one too? since it seems like it's a duplicate?

matthewcarlreetz (approved):

tyler-whitman:
> should we remove the other one too? since it seems like it's a duplicate?

Actually that one is being used in the ModalPicker component. So I guess we may want to choose one or the other for the future? @matthewcarlreetz any reason that we need both picker libraries?

tyler-whitman:
I'm going to merge this for now to fix the build and we can decide on one library for a future PR.

PR #114, opened 2021-01-21
Feature/hotspot settings

matthewcarlreetz:
On hotspot detail when pressing the hamburger menu an action sheet will show. Pressing the "settings" option will launch the diagnostic flow.

https://user-images.githubusercontent.com/1760821/105385440-f962d700-5bd8-11eb-8fdf-cc20fe4461cf.mov

### DIFF
--- a/src/features/hotspots/details/HotspotDetailCardHeader.tsx
+++ b/src/features/hotspots/details/HotspotDetailCardHeader.tsx
@@ -20,7 +20,7 @@ const HotspotDetailCardHeader = ({
         <Heart />
       </TouchableOpacityBox>
       <CardHandle />
-      <TouchableOpacityBox onPress={onMoreSelected}>
+      <TouchableOpacityBox onPress={onMoreSelected} padding="m" margin="n_m">
         <MoreMenu />
       </TouchableOpacityBox>
     </Box>
--- a/src/features/hotspots/details/HotspotDetailCardHeader.tsx
+++ b/src/features/hotspots/details/HotspotDetailCardHeader.tsx
@@ -14,16 +14,22 @@ const HotspotDetailCardHeader = ({
   onFollowSelected,
   onMoreSelected,
 }: Props) => (
-  <Box padding="m">
-    <Box flexDirection="row" justifyContent="space-between" padding="s">
-      <TouchableOpacityBox onPress={onFollowSelected}>
-        <Heart />
-      </TouchableOpacityBox>
-      <CardHandle />
-      <TouchableOpacityBox onPress={onMoreSelected} padding="m" margin="n_m">
-        <MoreMenu />
-      </TouchableOpacityBox>
-    </Box>
+  <Box flexDirection="row" justifyContent="space-between" paddingTop="m">
+    <TouchableOpacityBox
+      onPress={onFollowSelected}
+      paddingVertical="s"
+      paddingHorizontal="l"
+    >
+      <Heart />
+    </TouchableOpacityBox>
+    <CardHandle />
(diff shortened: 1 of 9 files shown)


src/features/hotspots/details/HotspotDetailCardHeader.tsx
```diff
@@ -20,7 +20,7 @@ const HotspotDetailCardHeader = ({
         <Heart />
       </TouchableOpacityBox>
       <CardHandle />
-      <TouchableOpacityBox onPress={onMoreSelected}>
+      <TouchableOpacityBox onPress={onMoreSelected} padding="m" margin="n_m">
```
allenan, line 23: just curious what the padding + negative margin does here?

src/features/hotspots/details/HotspotDetails.tsx
```diff
@@ -59,6 +57,31 @@ const HotspotDetails = () => {
   const { t } = useTranslation()
   const selectedHotspots = hotspotsToFeatures([hotspot])
   const { purpleMain, greenOnline } = useColors()
+  const { showActionSheetWithOptions } = useActionSheet()
+  const [showSettings, setShowSettings] = useState(false)
+
+  const onMoreMenuSelected = () => {
+    const options: string[] = [
+      ...t('hotspot_details.more', { returnObjects: true }),
```
allenan, line 65: returning an array here, and then making assumptions below based on the index feels a little implicit. would it be worth having this be a bit more explicit at the cost of being more verbose?

allenan: you could maybe create an object with labels and handler functions. would remove the need for the switch below, potentially

src/features/hotspots/root/HotspotsView.tsx
```diff
@@ -178,27 +175,15 @@ const HotspotsView = ({ ownedHotspots }: Props) => {
...
-          >
-            <Settings />
-          </TouchableOpacityBox>
-        )}
         initialSnapIndex={1}
         snapProgress={snapProgress}
       >
         <HotspotsCarousel
-          hotspots={ownedHotspots}
+          hotspots={[...ownedHotspots, ...ownedHotspots]}
```
allenan, line 182: is this maybe left in from testing?

src/features/hotspots/details/HotspotDetailCardHeader.tsx
matthewcarlreetz: It gives some hit slop. The button needed some padding, but the container had the padding already. This gives it the padding, then the n_m repositions it where it should be.

src/features/hotspots/details/HotspotDetails.tsx
matthewcarlreetz: yeah, I'm with you. It does feel a bit smelly.

src/features/hotspots/root/HotspotsView.tsx
matthewcarlreetz: damn. Yes.

src/features/hotspots/details/HotspotDetails.tsx
matthewcarlreetz: Your suggestion is similar to what I did for the more page. I'll go with that.

tyler-whitman (approved):
Looks good, just wanted to make sure the yarn changes are ok? Looks like a lot changed

allenan (approved):
awesome, send it!

DGKSK8LIFE (approved):
Lookin' good!

PR #120, opened 2021-01-28
add pending txn after send

allenan:
I added the same pattern of dispatching a pending txn, but I didn't see it show up right away. need to debug further, and also maybe associate this with realtime updates


matthewcarlreetz:
> I added the same pattern of dispatching a pending txn, but I didn't see it show up right away. need to debug further, and also maybe associate this with realtime updates

With regard to real-time updates... The more I think about it the more I like the idea of refreshing the data when the user changes tabs and/or brings the app into the foreground. Not sure if you've settled on a pattern, but maybe worth considering.

allenan:
> > I added the same pattern of dispatching a pending txn, but I didn't see it show up right away. need to debug further, and also maybe associate this with realtime updates
> 
> With regard to real-time updates... The more I think about it the more I like the idea of refreshing the data when the user changes tabs and/or brings the app into the foreground. Not sure if you've settled on a pattern, but maybe worth considering.

Yeah I ended up doing something like that in a local branch before getting distracted by the bottom sheet stuff. will get that checked in tomorrow

matthewcarlreetz:
> > > I added the same pattern of dispatching a pending txn, but I didn't see it show up right away. need to debug further, and also maybe associate this with realtime updates
> > 
> > 
> > With regard to real-time updates... The more I think about it the more I like the idea of refreshing the data when the user changes tabs and/or brings the app into the foreground. Not sure if you've settled on a pattern, but maybe worth considering.
> 
> Yeah I ended up doing something like that in a local branch before getting distracted by the bottom sheet stuff. will get that checked in tomorrow

There is a hook `useVisible` that tracks when a view when goes on/off the screen. Might be useful 🤷 
`src/utils/useVisible.ts`

PR #124, opened 2021-01-29
Transfer Hotspot

tyler-whitman:
Enables transfers of hotspots
- Initiate a transfer from hotspot settings. This send a notification to the buyer
- Buyer can access the transfer from their notifications and submit it
- Seller can cancel transfers
- Checks seller doesn't change the amount while buyer is submitting
- Checks for cancellations on buyers side
- Check for nonce issues on buyers side (buyer does a payment while the transfer is pending)

Currently known issues:
1. Hotspots don't refresh automatically after transfer (@allenan is working on making sure data doesnt get stale)
2. The pending transaction after you submit the transaction isnt working yet
3. Push notification are currently not working, but the notification will show in the list


allenan:
> * Enables transfers of hotspots
> 
> Currently two known issues:
> 
> 1. Hotspots don't refresh automatically after transfer (@allenan is working on making sure data doesnt get stale)
> 2. The pending transaction after you submit the transaction isnt working yet

yeah hoping to address both of those issues in my branch

src/features/hotspots/transfers/TransferRequests.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export type Transfer = {
+  amountToSeller?: Balance<NetworkTokens>
+  partialTransaction: string
+  gateway: string
+  seller: string
+  buyer: string
+}
+
+export const createTransfer = async (
```
allenan, line 24: not a big deal, but just wondering if this is the right place for this file? would it be more at home in `utils` or even as a redux slice?

src/features/hotspots/transfers/HotspotTransfer.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const hotspotName = animalName(hotspot.address)
+  const { t } = useTranslation()
+  const navigation = useNavigation()
+  const [typedName, setTypedName] = useState('')
+
+  const handleTypeName = (text: string) => {
+    setTypedName(text)
+  }
+
+  const typedHotspotName = () => {
```
allenan, line 34: i was confusing this with typescript types in my head for a moment. of course i can't immediately think of a better name off the top of my head 😛

src/features/wallet/send/SendView.tsx
```diff
@@ -142,7 +233,97 @@ const SendView = ({ scanResult }: { scanResult?: QrScanResult }) => {
     triggerNavHaptic()
   }
 
-  const constructTxn = async (): Promise<string> => {
+  const handleSellerTransfer = async (): Promise<undefined> => {
```
allenan, line 238: is `Promise<undefined>` required here? could probably leave it off and it'd be inferred from being an async function

allenan (approved):
code looks good. we can test the functionality once it's on testflight

src/features/hotspots/transfers/TransferRequests.tsx
tyler-whitman: Hmm yeah good point, I probably should move the data to redux as right now its integrated into the components. I think in regards to time I would leave it as it is for now. Most of the data is already in the SendComponent anyway. 

I kind of like having all the transfer stuff in one folder so its easier to find. I could move this to utils though?

src/features/hotspots/transfers/HotspotTransfer.tsx
tyler-whitman: Haha hmm, I could make it `enteredHotspotName`? Its the name they are typing to confirm the transfer

src/features/wallet/send/SendView.tsx
tyler-whitman: Good point, removed it

src/features/hotspots/transfers/TransferRequests.tsx
tyler-whitman: Going to merge it now so people can start testing and we can address this is a separate PR if we want to move it to redux

PR #127, opened 2021-01-30
Feature/update location

### LINKED ISSUE #93: Hotspot Settings > Update Location
An entry point for onboarded Hotspots to reassert their location.

It should work exactly as it does now.

Settings > Pair with Hotspot > Update Location

User will be presented with a screen to determine if the assert is free, or if they have to pay $10 in DC. Once they confirm, they submit the transaction. Go to wallet page.

- [ ] The Wallet page will have a pending assert transaction

matthewcarlreetz:
Closes #93


src/utils/useHotspot.ts
```diff
@@ -484,9 +481,23 @@ const useHotspot = () => {
...
       hasSufficientBalance,
       totalStakingAmount,
+      totalStakingAmountDC,
+      totalStakingAmountUsd,
+      remainingFreeAsserts: remainingFreeAsserts(),
     }
   }
 
+  const remainingFreeAsserts = () => {
+    if (!connectedHotspotDetails.validOnboarding) {
```
allenan, line 491: your other PR gets rid of `validOnboarding` right? might need to adjust this to check for onboarding record

matthewcarlreetz: ok, yeah, good call. Rebasing right now.

src/components/Map.tsx
```diff
@@ -72,27 +66,38 @@ const Map = ({
...
+      zoomLevel: userCoords ? 16 : 2,
       animationDuration: 500,
       heading: 0,
     })
   }
+  const flyTo = useCallback(
+    async (lat?: number, lng?: number, duration?: number) => {
+      if (!lat || !lng) return
+
+      await sleep(300) // this fixes a timing bug on Android where zoom level doesn't get set when flyTo is called to soon
```
tyler-whitman, line 81: Where is the call to change the zoom level? This happens when you have animations and the next camera call happens before the current one finishes its animation. The better thing to do is to do both the zoom level and the flyTo in the same call to the camera so that you wont run into this problem. 

EG use `camera.current?.setCamera` like:
```
    camera.current?.setCamera({
      centerCoordinate: currentLocation
        ? [currentLocation.longitude, currentLocation.latitude]
        : [-98.35, 15],
      zoomLevel: currentLocation ? 16 : 2,
      animationDuration: 500,
      heading: 0,
    })
```

src/components/Map.tsx
```diff
@@ -163,6 +167,28 @@ const Map = ({
         zoomEnabled={interactive}
         compassEnabled={false}
       >
+        {(showUserLocation || currentLocationEnabled) && (
+          <MapboxGL.UserLocation
+            onUpdate={(loc) => {
```
allenan, line 172: given that this callback gets called super frequently, we might want to move it into a defined const function with `useCallback`

src/features/hotspots/details/HotspotDetails.tsx
```diff
@@ -289,11 +290,13 @@ const HotspotDetails = () => {
         </BottomSheet>
       </Box>
 
-      <HotspotSettings
-        hotspot={hotspot}
-        visible={showSettings}
-        onClose={() => setShowSettings(false)}
-      />
+      <HotspotSettingsProvider>
```
allenan, line 293: we'll need to figure out the best way to integrate these settings provider changes with the bottom sheet branch.

src/features/hotspots/settings/HotspotSettingsProvider.tsx
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  const goBack = () => {
+    backHandler?.current?.()
+  }
+
+  return { showBack, enableBack, disableBack, goBack }
+}
+
+const HotspotSettingsContext = createContext<
```
allenan, line 36: one thing I was curious is how we're differentiating between what should be a context/provider vs a redux slice? In my branch, I started breaking some things out into more granular slices. One for activity etc. I could see this being a hotspot settings slice. is there any reason to pick provider/context over redux? or should we start trying to centralize on redux?

src/components/Button.tsx
```diff
@@ -15,31 +15,40 @@ type Props = BoxProps<Theme> & {
...
   textStyle,
   disabled,
+  height,
   ...rest
 }: Props) => {
   const getBackground = (): Colors | undefined => {
     if (mode === 'contained') {
-      if (variant === 'secondary') {
-        return 'secondaryMain'
+      switch (variant) {
```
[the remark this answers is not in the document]
matthewcarlreetz: Yep, good call. I like it.

src/components/Map.tsx
matthewcarlreetz: I went ahead with this. Works well 👍

src/features/hotspots/settings/HotspotSettingsProvider.tsx
matthewcarlreetz: I think you're right that we should centralize on redux by default. For this instance, I used a context because storing functions in redux is frowned upon. I'm sure it would work though. I probably could have architected this differently and used redux to track the overall navigation state for HotspotSettings. I took the approach of allowing each "sub" view to manage their own navigation.

allenan: ok that's fine, we'll just need to see how this plays with the changes i made in bottom sheet

allenan (approved):

PR #128, opened 2021-02-01
new bottom sheet component

allenan:
I switched our bottom sheet component to try [@gorhom/bottom-sheet](https://github.com/gorhom/react-native-bottom-sheet) and found it covered our use cases pretty well.

I had to restructure a number of things, including moving more things to redux, which is probably just good practice anyway.

### Todo:

- [x] Test Android

#### Hotspots Screen
- [x] show witnesses
- [x] show settings
- [ ] change handle to show options on details view?

#### Wallet Screen
- [x] fix snap points
- [x] fix card animation
- [x] use [bottom sheet modal](https://gorhom.github.io/react-native-bottom-sheet/modal) for txn details
- [x] add ref to snap sheet to bottom position on receive button press
- [x] filter transactions, move to redux


matthewcarlreetz:
Seems to work well. I tested on two different android phones and the touches are working great.

tyler-whitman:
Also tested on Android and iOS and its working well! Love the new hotspot details animation and shared map. I am seeing some warnings for `Calling %s on the ref of an Animated component is no longer necessary. You can now directly use the ref instead. This method will be removed in a future release. SectionList getNode()
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:285292:33`

allenan:
> Also tested on Android and iOS and its working well! Love the new hotspot details animation and shared map. I am seeing some warnings for `Calling getNode() on the ref of an Animated component is no longer necessary. You can now directly use the ref instead. This method will be removed in a future release.`

Yeah I see those as well. seems to be something within the bottom sheet component. I'm hoping to cut over to v3 of that library once it's a little more stable, but I can also see if there's an issue for that warning already and file one if not

allenan:
This is ready for review. I'm not sure why that test is failing. I spent some time trying to track down the issue and seems like something fairly new with the reanimated rc.2?

I also had to move the settings icon below the handle because it's kind of a pain to change the handle in that bottom sheet navigator. Another option would be to use the bottom sheet stacked modal for the hotspot list and the hotspot details, which would make them discrete cards with their own handle, but I just left it like this for right now. easy to change that up later

babel.config.js
```diff
@@ -1,12 +1,23 @@
...
 const baseConfig = {
   presets: ['module:metro-react-native-babel-preset'],
   plugins: [
     [
       'module-resolver',
       {
-        alias: {
-          '@helium/crypto': './node_modules/@helium/crypto-react-native',
-        },
+        alias: getAliasesFromTsConfig(),
```
allenan, line 18: this allows us to use `@assets/images/...etc` as an alias. I tried it with `@components` too but typescript wasn't able to track down their definitions, so just leaving it as assets for now

src/components/Map.tsx
```diff
@@ -20,9 +22,9 @@ type Props = {
   currentLocationEnabled?: boolean
   zoomLevel?: number
   mapCenter?: number[]
-  ownedHotspots?: Feature[]
-  selectedHotspots?: Feature[]
-  witnesses?: Feature[]
+  ownedHotspots?: Hotspot[]
```
allenan, line 25: I changed the interface for these to take `Hotspots` so that the places where we're calling `Map` don't need to do any geo-logic

src/features/wallet/root/ActivityCard/ActivityCardHeader.tsx
```diff
@@ -60,4 +68,4 @@ const ActivityCardHeader = ({ filter, onFilterChanged }: Props) => {
   )
 }
 
-export default memo(ActivityCardHeader)
```
allenan, line 63: what's the purpose of putting memo on every export? should we be doing this everywhere? is it necessary?

src/features/wallet/root/WalletScreen.tsx
```diff
@@ -12,15 +10,7 @@ const WalletScreen = () => {
         <WalletView />
       </Box>
 
-      {/* For some reason having a BottomSheet (which is in ActivityDetails)
-      contained in a <Portal/> breaks the header drag handler on Android.
-      Conditionally removing the portal on Android for now  */}
-      {Platform.OS === 'android' && <ActivityDetails />}
```
allenan, line 18: @matthewcarlreetz can you check that this part works ok on Android?

src/features/hotspots/root/HotspotsView.tsx
```diff
@@ -1,141 +1,123 @@
...
-        color="black"
-        paddingVertical="m"
-        paddingStart="l"
-      >
-        {t('hotspots.owned.your_hotspots')}
-      </Text>
-    </Box>
-  )
+  const snapPoints = useMemo(() => {
+    return selectedHotspot ? [0.1, 140, '80%'] : [0.1, '38%', '80%']
```
matthewcarlreetz, line 119: Cool that you can use percentages.

src/utils/appDataClient.ts
```diff
@@ -35,7 +36,7 @@ export const getHotspots = async () => {
   return newHotspotList.takeJSON(1000)
 }
 
-export const getHotspotDetails = async (address: string) => {
+export const getHotspotDetails = async (address: string): Promise<Hotspot> => {
```
matthewcarlreetz, line 39: Is it your preference to have explicit return types?

src/utils/useToggle.ts
```diff
@@ -0,0 +1,10 @@
+import { useCallback, useState } from 'react'
+
```
matthewcarlreetz, line 2: nice. This will come in handy.

src/store/activity/activitySlice.ts
```diff
@@ -0,0 +1,114 @@
+import { createSlice, createAsyncThunk, PayloadAction } from '@reduxjs/toolkit'
```
matthewcarlreetz, line 1: Yeah, good call on separating this out.

src/store/account/accountSlice.ts
```diff
@@ -1,14 +1,7 @@
-import { createSlice, createAsyncThunk, PayloadAction } from '@reduxjs/toolkit'
```
matthewcarlreetz, line 1: accountSlice has become a bit generic. We could probably take this further and separate out notifications. I'm wondering if we can remove `hotspots` now that we have `hotspotsSlice`?

src/features/wallet/scan/ScanView.tsx
```diff
@@ -110,68 +113,70 @@ const ScanView = () => {
...
             >
-              View my QR code
-            </Text>
-            <Text marginBottom="xs">
-              HNT can be burned into Data Credits to pay for device network
-              connectivity
-            </Text>
-            <Text variant="body2Bold" color="blueMain">
-              Learn More
+              Find a QR Code to scan
```
matthewcarlreetz, line 125: There are some missing translations here.

src/features/wallet/root/WalletScreen.tsx
matthewcarlreetz: yep, the code is looking good. I will do some thorough device testing next...

src/features/wallet/root/ActivityDetails/ActivityDetails.tsx
```diff
@@ -21,123 +19,84 @@ import HotspotTransaction from './HotspotTransaction'
...
 const DF = 'MM/dd/yyyy hh:mm a'
 const ActivityDetails = () => {
-  type BottomSheetHandle = ElementRef<typeof BottomSheet>
-  const sheet = useRef<BottomSheetHandle>(null)
+  const sheet = useRef<BottomSheetModal>(null)
   const { result: address } = useAsync(getSecureItem, ['address'])
   const { t } = useTranslation()
-  const { activityItem, setActivityItem } = useWalletContext()
+  const dispatch = useAppDispatch()
+  const {
```
matthewcarlreetz, line 34: Moving to redux is a good improvement 💯

src/features/wallet/root/WalletScreen.tsx
matthewcarlreetz: It looks like this still is an issue. The drag handle on android doesn't work for the detail view. Clicking the background to close does work though.

src/features/wallet/root/ActivityCard/ActivityCardHeader.tsx
```diff
@@ -1,27 +1,35 @@
...
   const { t } = useTranslation()
+  const dispatch = useAppDispatch()
+  const {
+    activity: { filter },
+  } = useSelector((state: RootState) => state)
+
   const filters = t('transactions.filter', { returnObjects: true }) as Record<
     string,
     string
   >
```
matthewcarlreetz, line 28: The filter isn't working right on Android. I don't think it's related to your PR though. It may have been a miss in my testing when building `<ModalPicker>` component.

src/features/wallet/scan/ScanView.tsx
allenan: yeah these are still just yolo placeholders. I just changed the bottom sheet wrapping. Will move to translations once this is finalized

allenan:
> > This is ready for review. I'm not sure why that test is failing. I spent some time trying to track down the issue and seems like something fairly new with the reanimated rc.2?
> > I also had to move the settings icon below the handle because it's kind of a pain to change the handle in that bottom sheet navigator. Another option would be to use the bottom sheet stacked modal for the hotspot list and the hotspot details, which would make them discrete cards with their own handle, but I just left it like this for right now. easy to change that up later
> 
> yeah, I actually decided not to upgrade to rc2 because of this. I can dig into a fix a little more today.

hm yeah I think other things break on the tests with rc1 + bottom sheet, so might as well go with the newer version

src/store/account/accountSlice.ts
allenan: yeah I was thinking we could work towards that. just didn't want to bite off too much here

babel.config.js
tyler-whitman: Did this remove the `@helium/crypto` one though?

matthewcarlreetz (approved):
🚀

PR #169, opened 2021-02-07
Feature/hotspots stacked modals

allenan:
Use stacked modal sheets on the hotspots screen rather than a sheet navigator. First pass, might still need some optimization. And haven't investigated a carousel inside the details modal yet


index.js
```diff
@@ -1,3 +1,4 @@
+import './src/whyDidYouRender'
```
tyler-whitman, line 1: Should we import this only for `DEV`?

src/components/EmojiBlip.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const emojiOptions = [...emojis.anytime]
+  const hour = date.getHours()
+
+  if (hour >= 4 && hour < 12) {
+    emojiOptions.push(...emojis.morning)
+  }
+  if (hour >= 12 && hour < 15) {
+    emojiOptions.push(...emojis.lunch)
+  }
+  if (hour >= 18 || hour < 4) {
```
tyler-whitman, line 34: I did `hours >= 17 || hours < 4` in the `TimeOfDayTitle` maybe you can change that to 18 or change this to 17 so they line up?

src/features/hotspots/details/HotspotDetailCardHeader.tsx
```diff
@@ -1,36 +1,77 @@
...
+  const dispatch = useAppDispatch()
+  const { showActionSheetWithOptions } = useActionSheet()
+
+  const {
+    hotspotDetails: { hotspot },
+  } = useSelector((state: RootState) => state)
+
+  type SettingsOption = { label: string; action?: () => void }
+  const onMoreSelected = () => {
+    if (!hotspot) return
```
tyler-whitman, line 26: This click isnt working for me, for some reason hotspot is always undefined

src/components/EmojiBlip.tsx
allenan: yeah I had that but then most of my afternoon emojis ended up being food related so wanted it closer to lunch time 🙃

src/features/hotspots/details/HotspotDetailCardHeader.tsx
allenan: hm I think it's because that card header doesn't get rerendered when the hotspot changes, probably for good reason. I think I'll need to move that down below the handle

src/components/EmojiBlip.tsx
tyler-whitman: Ok haha, sounds good then

tyler-whitman (approved):

PR #172, opened 2021-02-08
Show security screen when app is inactive.

### LINKED ISSUE #139: Security screen
Like with the current app, we should hide the context switcher preview with the splash screen when the app is in the background

matthewcarlreetz:
Closes #139


src/App.tsx
```diff
@@ -10,6 +9,8 @@ import {
   UIManager,
   LogBox,
 } from 'react-native'
+import { ThemeProvider } from '@shopify/restyle'
+
```
tyler-whitman, line 13: Nit picking, but we can remove the space haha

src/features/security/SecurityScreen.tsx
```diff
@@ -0,0 +1,65 @@
...
+type Props = { visible: boolean }
+const SecurityScreen = ({ visible }: Props) => {
+  const fadeAnim = useRef(new Animated.Value(0))
+  const [mounted, setMounted] = useState(false)
+
+  const fadeIn = () => {
+    setMounted(true)
+    Animated.timing(fadeAnim.current, {
+      toValue: 1,
+      duration: 500,
```
tyler-whitman, line 15: I think this needs to be a little faster. Sometimes when I minimize on iOS I can still see the content and the security screen is slightly faded above it. When I set it to `100` it seems to work

tyler-whitman: I think whats happening is maybe the animation stops as it goes to background?

tyler-whitman: Hmm it also might be the map slowing down the animation thread? I think it happens on when I am also still loading the map. Might be a debug only issue

tyler-whitman: Hmm also doesnt seem to be working on Android. But the visible boolean is getting set correctly. My android device is pretty slow though.. but I'm not seeing it ever appear

matthewcarlreetz: You're right about Android. It appears to not be working in the old app either. I'll investigate.

tyler-whitman (approved):
Nice, working well on Android now too

src/features/security/SecurityScreen.tsx
tyler-whitman: Fixed with your new changes

PR #183, opened 2021-02-10
Remove prompt to add hotspot after onboarding.

### LINKED ISSUE #144: After account recovery, do not immediately prompt user to add hotspot.
There is inherently some uncertainty when setting up the app from seed words. The first screen I saw after the post-seed-word entry "Recovering" spinner screen was the "Add Hotspot" screen (attached), which led me to initially think I'd entered my seed words incorrectly and a new wallet had been created.

Ideally I would be taken directly to the "Your Hotspots" screen where I would see that my hotspots had been imported. Even better would be a nice confirmation that my account had been imported successfully.

<kbd></kbd>

matthewcarlreetz:
Closes #144


tyler-whitman (approved):
Looks good

allenan:
I came across this unrelated bug: https://github.com/helium/hotspot-app/issues/190

allenan (approved):

PR #186, opened 2021-02-11
Feature/activity loading ux

### LINKED ISSUE #182: Improve Activity List Loading UX
The activity list loading is currently pretty jarring. A few ideas for improving the behavior...

Use skeleton ui when initially loading or switching filter
https://github.com/chramos/react-native-skeleton-placeholder#readme

When list auto-refreshes show a loading indicator outside of the list. This way the ui doesn't jump around when the new list likely has the same results.

matthewcarlreetz:
Closes #182 

Adds a skeleton view when initial activity data is loading. Removes jarring ui transitions when activity data is loading.


allenan:
noticed the spacing on the arrow seems off for the "all activity" state:

image

src/features/wallet/root/ActivityCard/ActivityCard.tsx
```diff
@@ -59,17 +58,21 @@ const ActivityCard = forwardRef((props: Props, ref: Ref<BottomSheet>) => {
...
+      data = pendingTxns
     }
+
     if (filter === 'all') {
-      return [...pendingTxns, ...txns]
+      data = [...pendingTxns, ...txns]
+    }
+
+    if (showSkeleton) {
+      return skeletonData(data.length)
```
allenan, line 72: this is working great right now but one thought I had, would it be cleaner to cut over to an entirely different component with some fixed number of items in it for the skeleton view? that way we're not interweaving skeleton logic throughout our activity fetching code?

matthewcarlreetz: Yeah, I definitely hear what you're saying. That likely would be cleaner. I'll give it a try and see how it feels.

allenan (approved):

PR #189, opened 2021-02-11
fix stacked modal layout issue on small phones

allenan:


tyler-whitman (approved):
Working well on iPhone X

PR #192, opened 2021-02-12
Bugfix/hotspot onboarding

matthewcarlreetz:
During hotspot onboarding, it was possible to double click any of the buttons which would cause various errors along the way.

This PR creates a higher-order component that debounces button clicks. I've created "debounced" versions of our buttons and converted the hotspot onboarding process to use them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,7 @@
     "react-redux": "^7.2.2",
     "redux-thunk": "^2.3.0",
     "rn-sliding-up-panel": "^2.4.4",
+    "use-debounce": "^5.2.0",
     "use-state-with-callback": "^2.0.3",
     "validator": "^13.5.1"
   },
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -7,6 +7,7 @@ import Text from './Text'
 import { Colors, Theme } from '../theme/theme'
 import TouchableOpacityBox from './TouchableOpacityBox'
 import Box from './Box'
+import WithDebounce from './WithDebounce'
 
 type Props = BoxProps<Theme> & {
   mode?: 'text' | 'contained'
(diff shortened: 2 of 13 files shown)


allenan:
I think debouncing is a good strategy in general -- we may want to just make that a default behavior of our main button/touchable box components. I don't see a case where we wouldn't want it. Additionally, I'm wondering if during onboarding or any case where destructive things are happening if we should use the strategy of disabling the button until the task completes or fails? I'm pretty sure we do that in some places, but I'm sure it could be added elsewhere.

matthewcarlreetz:
> I think debouncing is a good strategy in general -- we may want to just make that a default behavior of our main button/touchable box components. I don't see a case where we wouldn't want it. Additionally, I'm wondering if during onboarding or any case where destructive things are happening if we should use the strategy of disabling the button until the task completes or fails? I'm pretty sure we do that in some places, but I'm sure it could be added elsewhere.

I agree about using the debounce as the default. I just think we'd need to thoroughly test the app after making the cutover. I'd prefer to merge this PR to see if it helps Nebra, then convert all buttons in a separate PR.

allenan (approved):

allenan:
> > I think debouncing is a good strategy in general -- we may want to just make that a default behavior of our main button/touchable box components. I don't see a case where we wouldn't want it. Additionally, I'm wondering if during onboarding or any case where destructive things are happening if we should use the strategy of disabling the button until the task completes or fails? I'm pretty sure we do that in some places, but I'm sure it could be added elsewhere.
> 
> I agree about using the debounce as the default. I just think we'd need to thoroughly test the app after making the cutover. I'd prefer to merge this PR to see if it helps Nebra, then convert all buttons in a separate PR.

fair enough

PR #193, opened 2021-02-12
Feature/languages

### LINKED ISSUE #178: expose language picker in settings
we will have translations for Korean, Japanese, and Simplified Chinese. FR and ES are not translated, so those should be left out. I'm guessing the .ts files need to be remove from the locales folder?

matthewcarlreetz:
Closes #178

### DIFF
--- a/src/features/moreTab/more/MoreScreen.tsx
+++ b/src/features/moreTab/more/MoreScreen.tsx
@@ -28,7 +28,7 @@ import DiscordItem from './DiscordItem'
 import AppInfoItem from './AppInfoItem'
 import activitySlice from '../../../store/activity/activitySlice'
 import hotspotsSlice from '../../../store/hotspots/hotspotsSlice'
-import { useLanguage } from '../../../utils/i18n'
+import { SUPPORTED_LANGUAGUES, useLanguage } from '../../../utils/i18n'
 
 type Route = RouteProp<RootStackParamList & MoreStackParamList, 'MoreScreen'>
 const MoreScreen = () => {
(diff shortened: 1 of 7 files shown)


src/features/moreTab/more/MoreScreen.tsx
```diff
@@ -128,6 +129,13 @@ const MoreScreen = () => {
     )
   }, [t, dispatch])
 
+  const handleLanguageChange = useCallback(
+    (lng: string) => {
```
allenan, line 133: this is sort of silly but maybe we could name this `lang` or `language`? we use `lng` pretty extensively to mean `longitude` elsewhere :)

src/locales/ja.ts
```diff
@@ -1,3 +1,3 @@
 export default {
-  back: 'Retour',
+  back: 'jBack',
```
allenan, line 2: is this change intentional?

src/features/moreTab/more/MoreScreen.tsx
matthewcarlreetz: yeah, I was thinking the same thing, but i18n uses `lng` so I stuck with it. I'll change it up.

src/locales/ja.ts
matthewcarlreetz: I mean, sort of 🤷 

I renamed the fr file to ja and just threw that in there for the moment. Should I copy over the English strings for now?

allenan: oh i see. i guess i would delete anything that's not currently being used and let it fall back to the english translation

matthewcarlreetz: I removed the key for now

allenan:
is there a setting (maybe just in release builds?) to have strings without a translation in a given language fallback to english, rather than the i18n string?

image

allenan (approved):
works really well!

PR #194, opened 2021-02-13
Hotspot Checklist

### LINKED ISSUE #92: Hotspot Checklist
Bring in the Hotspot Checklist from Explorer to the new mobile app.
Design files are in Figma > Hotspot Checklist

image

tyler-whitman:
closes #92 

For now we are not showing the checklist progress on the hotspots list, we would probably need to add the data to wallet api as it will be too many calls for each hotspot. 

loading

collapsed

complete

in progress


src/assets/images/carot-down-picker.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="8" height="6" viewBox="0 0 8 6" fill="none" xmlns="http://www.w3.org/2000/svg">
```
allenan, line 1: is this asset a duplicate of the one below it?

src/features/hotspots/checklist/HotspotChecklist.tsx
```diff
@@ -0,0 +1,267 @@
...
+      loadingActivity,
+    },
+  } = useSelector((state: RootState) => state)
+
+  useEffect(() => {
+    dispatch(fetchChecklistActivity(hotspot.address))
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  const syncStatus = () => {
```
allenan, line 56: this const and the ones below could be wrapped in `useMemo`

src/features/hotspots/checklist/HotspotChecklist.tsx
```diff
@@ -0,0 +1,267 @@
...
+  checklistData.sort((a, b) => Number(b.complete) - Number(a.complete))
+  const firstIndex = checklistData.findIndex((i) => !i.complete)
+  const [hidden, setHidden] = useState(true)
+
+  const toggleHidden = () => {
+    animateTransition()
+    setHidden(!hidden)
+  }
+
+  const renderItem = ({ item }: { item: ChecklistItem }) => (
```
allenan, line 184: this could be wrapped in `useCallback`

src/features/hotspots/checklist/HotspotChecklist.tsx
```diff
@@ -0,0 +1,267 @@
...
+          sliderWidth={wp(100)}
+          itemWidth={wp(90)}
+          inactiveSlideScale={1}
+        />
+      )}
+    </Box>
+  )
+}
+
+export default HotspotChecklist
```
allenan, line 267: could add `memo` here

src/assets/images/carot-down-picker.svg
tyler-whitman: Yeah, I couldn't get it to work in the picker with a custom color so I just made one specifically for the picker since we use that color in multiple places and then I made another where you can change the color. Not sure if there is a better way?

src/features/hotspots/checklist/HotspotChecklist.tsx
tyler-whitman: Good point, updating

src/features/hotspots/checklist/HotspotChecklist.tsx
tyler-whitman: done

src/features/hotspots/checklist/HotspotChecklist.tsx
tyler-whitman: done

src/assets/images/carot-down-picker.svg
allenan: ah ok yeah idk, might have to change something about the picker. this is fine

allenan (approved):
looks great

PR #198, opened 2021-02-16
Remove animation pandemonium

### LINKED ISSUE #195: white flash when switching tabs sometimes
Based on some git checkouts, I believe `1852975 Feature/activity loading ux (#186)` introduced this.

matthewcarlreetz:
Fixes white flash when navigating away from activity tab. I believe the root cause was that there were overlapping calls to `LayoutAnimation.configureNext`

Closes #195

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -36,14 +36,11 @@ import {
 } from './store/helium/heliumDataSlice'
 import sleep from './utils/sleep'
 import SecurityScreen from './features/security/SecurityScreen'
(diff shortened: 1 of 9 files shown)


matthewcarlreetz:
> confirmed that this fixes the white flash, however I'm noticing a half-second or so delay when navigating off of the wallet tab to another tab

I see what you're talking about. I'll track it down.

allenan (approved):
much better now!

Issue #136, opened 2021-02-01
Fix Links in Settings > Learn


cokes518:

cokes518:
When new dev docs is migrated, update the old links.

PR #199, opened 2021-02-16
qr scan copy, learn link copy, externlize for translation

cokes518:


cokes518:
closes #136

src/features/wallet/scan/ScanView.tsx
```diff
@@ -122,7 +123,7 @@ const ScanView = () => {
               fontSize={18}
               marginBottom="s"
             >
-              Find a QR Code to scan
+              {t('send.scan.title')}
```
allenan, line 126: you did the externalizing too 😍

src/locales/en.ts
```diff
@@ -422,6 +422,16 @@ export default {
...
     last_activity: 'LAST REPORTED ACTIVITY: {{activity}}',
     label_error: 'You do not have enough HNT in your account.',
+    scan: {
+      title: 'Find a QR Code to Scan',
+      send: 'Send HNT',
+      send_description: 'Quickly scan a Helium address to send HNT.',
+      burn: 'Burn HNT to DC',
+      burn_description: 'HNT can be burned into Data Credits to pay for device network connectivity. DCs are non-transferable.',
+      view: 'View My QR Code',
+      view_description: 'Save your QR code for others to save and send payments.',
```
allenan, line 432: I was planning on having this slide up a modal that shows your own QR code. like if you hit the QR code button thinking that it would show yours rather than scan one. I don't know if that changes your description here or not

cokes518: Hm, I accessed this UI after hitting Send > QR icon, so that's where the copy's context is coming from. Is there another entry point?

src/features/wallet/scan/ScanView.tsx
cokes518: the best junior react engineer

src/locales/en.ts
allenan: the main entry point for your qr code is under the activity list when you press "receive". But figured this would be a good secondary entry point. Mainly I was just thinking rather than "Save" it could be like "Show your QR code to receive payments from others" or something? And rather than "Learn more" it could be "View QR Code" to launch a modal with it?

src/features/wallet/scan/ScanView.tsx
allenan: in addition to `import { useTranslation } from 'react-i18next'` you'll need to define `t`. At the top of this component, add `const { t } = useTranslation()`

src/locales/en.ts
```diff
@@ -422,6 +422,16 @@ export default {
...
     last_activity: 'LAST REPORTED ACTIVITY: {{activity}}',
     label_error: 'You do not have enough HNT in your account.',
+    scan: {
+      title: 'Ways to use a QR Code',
+      send: 'Send HNT',
+      send_description: 'Quickly scan a Helium address to send HNT.',
+      burn: 'Burn HNT to DC',
+      burn_description: 'HNT can be burned into Data Credits to pay for device network connectivity. DCs are non-transferable.',
+      view: 'View QR Code',
+      view_description: 'Send the QR Code as an easy way to populate a transaction.',
```
allenan, line 432: what about something like

```suggestion
      view_description: 'Share your QR Code to deposit or receive HNT from others.',
```

allenan (approved):

PR #207, opened 2021-02-17
Add support for ENS addresses

allenan:
[ENS](https://ens.domains/) supports HNT, this allows a valid ENS address to be used to send HNT.

Some unrelated code style changes also seem to have gotten picked up by my husky hook.

### DIFF
--- a/src/utils/explorerClient.ts
+++ b/src/utils/explorerClient.ts
@@ -1,0 +1,45 @@
+/* eslint-disable import/prefer-default-export */
+import * as Logger from './logger'
+
+const makeRequest = async (url: string, opts: RequestInit) => {
+  Logger.breadcrumb(`request: ${opts.method} ${url}`)
+  try {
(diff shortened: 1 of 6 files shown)


src/utils/explorerClient.ts
```diff
@@ -0,0 +1,45 @@
+/* eslint-disable import/prefer-default-export */
+import * as Logger from './logger'
+
+const makeRequest = async (url: string, opts: RequestInit) => {
+  Logger.breadcrumb(`request: ${opts.method} ${url}`)
+  try {
+    const route = ['https://explorer.helium.com/api', url].join('/')
```
matthewcarlreetz, line 7: Should we add this to the .env?

matthewcarlreetz (approved):

src/utils/explorerClient.ts
allenan: we could it just seemed like extra work to set it up in appcenter for something that wasn't private. I'll just leave it for now

Issue #203, opened 2021-02-17
Difficult to go 'Back' on My Hotspots view.


jthiller:
The placement of the back button makes it difficult to inspect more than one or two hotspots.

I took a quick pass at _one_ solution. I don't think this will be obvious to all users, but this pattern does exist in other apps. (e.g. in the Facebook app, pressing the active button after scrolling will return the user to the top of their feed)
https://www.figma.com/file/ZgcLuizxLDNExuBuF7htj8/Homescreen-Back?node-id=0%3A1

<img width="1726" alt="image" src="https://user-images.githubusercontent.com/1965053/108163367-bf83d400-70a3-11eb-8d11-5fe4b72f6335.png">

PeterMain:
+1 

I've hit that nav item a few times expecting it to return me home.

PR #220, opened 2021-02-18
Use tab button press to navigate back to hotspots list

allenan:
resolves https://github.com/helium/hotspot-app/issues/203


tyler-whitman (approved):
I like it, works as expected.

PR #226, opened 2021-02-18
Fix a couple pin entry edge cases.

### LINKED ISSUE #190: Account flow PIN/notifications issue
During the account onboarding flow, after setting your PIN, we prompt you for notification permissions.

1) we probably don't want to pop that without asking first?
2) regardless, because you've set a PIN, when the system modal pops, it puts the app into background triggering PIN reauth right after you've set it which feels odd.

Is there a way to prevent auth from triggering on those system notification pops?

matthewcarlreetz:
Closes #190 

User should no longer be asked for pin immediately after onboarding.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -103,7 +103,9 @@ const App = () => {
   )
 
   const loadInitialData = useCallback(async () => {
-    dispatch(restoreUser())
+    if (!isRestored) {
+      dispatch(restoreUser())
+    }
     dispatch(fetchInitialData())
 
     if (!isRestored && !isBackedUp) {
--- a/src/components/ConfirmPinView.tsx
+++ b/src/components/ConfirmPinView.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useRef, useEffect, useCallback } from 'react'
+import React, { useState, useRef, useEffect, useCallback, memo } from 'react'
 import { Animated } from 'react-native'
 import { useNavigation } from '@react-navigation/native'
 import Text from './Text'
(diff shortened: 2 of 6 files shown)


tyler-whitman (approved):

allenan:
So this would fix the issue assuming that they press the system alert within 5 seconds right? Does this also cover the screen with the security screen?

One thing I was thinking, I think there's an intermediate app state before "background", "inactive" I think. If that's the state the app is in while the system prompt is up, maybe we could be more specific about only triggering the security screen and pin auth when the state === "background"?

matthewcarlreetz:
> So this would fix the issue assuming that they press the system alert within 5 seconds right? Does this also cover the screen with the security screen?
> 
> One thing I was thinking, I think there's an intermediate app state before "background", "inactive" I think. If that's the state the app is in while the system prompt is up, maybe we could be more specific about only triggering the security screen and pin auth when the state === "background"?

There's a hook `src/utils/usePermissionManager.ts` that updates the flag `isRequestingPermission` in `appSlice.ts`. That flag is used to help determine if the lock screen is shown

`const shouldLock =
      isActive && isPinRequired && !isRequestingPermission && lastIdleExpired` // `App.tsx` line 89

I believe that is working well. The 5s delay was to help deal with FaceId on Android. LocalAuthentication puts the app in background on Android, but not on iOS.

I'll do some more thinking on this and see if I can find a better overall solution.

matthewcarlreetz:
@allenan I've done a bit of testing and it looks like android the only android states I see are `background` and `active`, so I think that leaves me having to track why the app went into background.

matthewcarlreetz:
Yeah, looks like `inactive` is an iOS only state

PR #227, opened 2021-02-19
Hotspot No Location View

### LINKED ISSUE #156: UI for Hotspots without location asserted
We're missing designs for Hotspots that have no location. Needs to be implemented.

tyler-whitman:
The mockups also had an assert button but I wasn't sure if that would be possible since we need to connect first. I could add a button and make it open the more menu or just leave it as is for now?

I also removed hotspots with no location from the map.

closes #156 

Screen Shot 2021-02-18 at 4 01 40 PM


src/features/hotspots/root/HotspotsView.tsx
```diff
@@ -171,7 +174,9 @@ const HotspotsView = ({ ownedHotspots }: Props) => {
       >
         <Map
           ownedHotspots={ownedHotspots}
-          selectedHotspots={selectedHotspot ? [selectedHotspot] : undefined}
+          selectedHotspots={
+            selectedHotspot && hasLocation ? [selectedHotspot] : undefined
```
matthewcarlreetz, line 178: I believe the `[selectedHotspot]` will be a different object on every render. Might want to memoize it.

matthewcarlreetz (approved):
Looks good. Just one little issue you might want to address.

allenan (approved):

PR #256, opened 2021-02-20
fixes reassert and assert location

allenan:
This fixes a couple issues:
- needed to increment nonce by 1
- if past the "free" limit, skip the onboarding server part and let the wallet pay directly
- make pending assert location txns show the mocked data and don't crash


matthewcarlreetz (approved):
Looks good to me. There's some TS stuff that could be cleaned up, but we can do that some other day.

src/utils/walletClient.ts
```diff
@@ -7,6 +7,7 @@ const makeRequest = async (url: string, opts: RequestInit) => {
   Logger.breadcrumb(`request: ${opts.method} ${url}`)
   try {
     const token = await getWalletApiToken()
+    // const token = undefined
```
tyler-whitman, line 10: Can remove this

tyler-whitman (approved):
Code looks good, not sure if I can test it haha

PR #258, opened 2021-02-21
Fix/misc fixes

### LINKED ISSUE #122: Design Tweaks: Wallet Page
Details are in Figma > Hotspot Tab (at the top), also pasted below.

image
image
image

### LINKED ISSUE #188: Fix gesture handling on charts in listview
when you pan down in the hotspot details list view, the charts often steal your gesture. We should tune that gesture handler so that it's not as sensitive to vertical gestures.

allenan:
resolves #122 
- sets wallet button icons to use a fixed height rather than width so that the icons will render the same size
- fixes security token amount display

resolves #188 
- tune the chart pan responder to reject gestures that are too vertical, this allows you to swipe down the hotspot charts view without having your gesture stolen by the chart

this isn't 100% perfect, you can still get a chart stuck in an active state if you manage to trigger the scroll gesture while the chart is active. But it doesn't impact the usability because you can still scroll and it's easy to get the chart unstuck. I think we could do something fancy with nested gesture-handlers and `waitFor`, but this is an improvement over the current situation.

### DIFF
--- a/src/features/wallet/root/BalanceCard/BalanceCard.tsx
+++ b/src/features/wallet/root/BalanceCard/BalanceCard.tsx
@@ -83,11 +83,7 @@ const BalanceCard = ({ onReceivePress, onSendPress }: Props) => {
           />
           <CurrencyBadge
             variant="hst"
-            amount={
-              account?.secBalance
-                ? parseFloat(account.secBalance.toString())
-                : 0
-            }
+            amount={account?.secBalance?.floatBalance || 0}
           />
         </Box>
       </Box>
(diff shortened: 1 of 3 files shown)


src/features/wallet/root/BalanceCard/BalanceCard.tsx
```diff
@@ -83,7 +83,11 @@ const BalanceCard = ({ onReceivePress, onSendPress }: Props) => {
           />
           <CurrencyBadge
             variant="hst"
-            amount={account?.secBalance?.integerBalance || 0}
+            amount={
```
tyler-whitman, line 86: Not sure what was wrong here, but maybe https://github.com/helium/helium-js/pull/146 would help too

tyler-whitman (approved):
Looks good

src/features/wallet/root/BalanceCard/BalanceCard.tsx
allenan: basically we needed the float balance instead of the integer balance. I just updated to a better way to do that

PR #261, opened 2021-02-21
Number locale formatting

### LINKED ISSUE #250: Non US Locale number formatting breaks send amount
Formatting issues with `,` and `.` for different locales are not working correctly.

tyler-whitman:
Changes number formatting throughout the app to support different locales.
Amount input now supports both formats as well and has a new fixed formatting input to help with pasting different formats in.
You can change your Region setting in the Language & Region iOS settings to test it out.

closes #250

Screen Shot 2021-02-20 at 9 18 44 PM
Screen Shot 2021-02-20 at 11 51 04 PM
Screen Shot 2021-02-20 at 11 51 31 PM


src/components/HotspotListItem.tsx
```diff
@@ -53,7 +54,10 @@ const HotspotListItem = ({
             </Text>
           </Box>
           <Text variant="body2" color="purpleMain" paddingTop="s">
-            {`+${totalReward.toString(2)}`}
+            {`+${totalReward.toString(2, {
```
allenan, line 57: could maybe encapsulate the i18n part of this into the currency util file?

src/components/InputField.tsx
```diff
@@ -57,30 +67,57 @@ const InputField = ({
...
+              precision={numberPrecision}
+              ignoreNegative
+              blurOnSubmit
+              autoCompleteType="off"
+              textContentType="none"
+              autoCapitalize="none"
+              autoCorrect={false}
+              dataDetectorTypes="none"
+              keyboardAppearance="dark"
+              style={{
```
allenan, line 87: we should probably be memozing styles

src/features/hotspots/details/HotspotDetailChart.tsx
```diff
@@ -33,7 +34,7 @@ const PercentageBox = ({
       {t('hotspot_details.pass_rate')}
     </Text>
     <Text variant="light" fontSize={32} color="black" marginBottom="s">
-      {`${focusedData ? focusedData.up : percentage}%`}
+      {`${focusedData ? focusedData.up.toLocaleString(locale) : percentage}%`}
```
allenan, line 37: could possible make a helper for localized percentages as well

allenan:
this looks really good and thorough! just wondering if there are any opportunities to compartmentalize some of the localization logic into reusable helpers so that it's easier to default to these versions going forward?

allenan:
Just trying this out, I don't think the send input feels very natural. It seems weird to start from the smallest decimal. Like to send 10 HNT I'd need to press `1` `0` `0` `0` `0` `0` `0` `0` `0` `0`. So I think that interaction will need some retooling

tyler-whitman:
> this looks really good and thorough! just wondering if there are any opportunities to compartmentalize some of the localization logic into reusable helpers so that it's easier to default to these versions going forward?

Hmm yeah was trying to thing of a way to do this and ended up with passing these params. Its tough because sometimes we need an actual toLocaleString() with the device local and sometimes we need to convert balance numbers which takes the separators. Any ideas to make it better?

> Just trying this out, I don't think the send input feels very natural. It seems weird to start from the smallest decimal. Like to send 10 HNT I'd need to press 1 0 0 0 0 0 0 0 0 0. So I think that interaction will need some retooling

I fixed this now and it works as it did before

src/components/HotspotListItem.tsx
tyler-whitman: Yeah its just that we do use other options too, like sometimes we want 8 decimals and sometimes we wanted fixed 2 or just rounded 2. Its a little tough to consolidate it more without ending up with something pretty similar

allenan (approved):
works great now!

PR #264, opened 2021-02-21
add details to hotspot list item

### LINKED ISSUE #260: Improve list view hotspot details
right now we simply show hotspot name, 24 hour earnings, and a static blue checkmark. we had to remove checklist preview due to api limitations.

Other details that would be helpful to see here:
- online/offline/needs-attention status
- location/city name
- witness count

this was a quick design sketch
image

allenan:
resolves #260 

just adding a different icon for a "needs attention" hotspot and geolocation data for now. If we wanted to do witness counts etc that'd probably involve restructuring data fetching logic, so just added the simpler things for now, and we can get user feedback on what else would be valuable to show here before doing a bunch of work.


tyler-whitman (approved):
Looks good

matthewcarlreetz (approved):

PR #268, opened 2021-02-21
Fix mnemonic not displaying after upgrading to v3.

### LINKED ISSUE #249: Unable to view 12 words from settings.
iPhone 12 Pro
iOS 14.4

To replicate:
- Open settings
- Tap 'reveal words' from Security section
- Validate with Face ID

matthewcarlreetz:
Closes #249 
Husky reformatted some stuff on me.

### DIFF
--- a/src/utils/secureAccount.ts
+++ b/src/utils/secureAccount.ts
@@ -75,7 +75,12 @@ export const getMnemonic = async (): Promise<Mnemonic | undefined> => {
   const wordsStr = await getSecureItem('mnemonic')
   if (!wordsStr) return
 
-  const words = JSON.parse(wordsStr)
+  let words: string[] = []
+  try {
+    words = JSON.parse(wordsStr) // The new (v3) app uses JSON.stringify ['hello', 'one', 'two', 'etc'] => "[\"hello\",\"one\",\"two\",\"etc\"]"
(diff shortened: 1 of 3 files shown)


src/utils/secureAccount.ts
```diff
@@ -75,7 +75,12 @@ export const getMnemonic = async (): Promise<Mnemonic | undefined> => {
   const wordsStr = await getSecureItem('mnemonic')
   if (!wordsStr) return
 
-  const words = JSON.parse(wordsStr)
+  let words: string[] = []
+  try {
+    words = JSON.parse(wordsStr) // The new (v3) app uses JSON.stringify ['hello', 'one', 'two', 'etc'] => "[\"hello\",\"one\",\"two\",\"etc\"]"
+  } catch (e) {
+    words = wordsStr.split(' ') // The old (v2) app space separated "hello one two etc"
```
allenan, line 82: should we "upgrade" v2 app users to the new format? or just support both?

allenan (approved):

src/utils/secureAccount.ts
matthewcarlreetz: I can definitely get onboard with that idea. I’ll get it in there today yet.

PR #269, opened 2021-02-22
Fix hotspot checklist not scrolling on android.

matthewcarlreetz:


allenan (approved):
tested ios, still works as expected there. seems like a valid approach for android 👍

PR #276, opened 2021-02-23
update hotspots list for offline hotspots

### LINKED ISSUE #253: Homescreen selecting offline hotspots
When on the home screen and  selecting your "offline hotspots" the list doesn't empty making it look like it's saying all of my hotspots are offline

### LINKED ISSUE #262: Homepage hotspot filter remembering state but clearing filter
When filtering hotspots by "top-performing hotspots" and then minimising the app and then coming back to it, the list is filtered by newest, but the previous selection "top-performing hotspots" is still chosen.

tyler-whitman:
- show "No Hotspots Offline" text for offline filter without offline hotspots
- ensure the list keeps the same filter if you leave the hotspots tab and come back

closes #253 
closes #262

### DIFF
--- a/src/components/HotspotListItem.tsx
+++ b/src/components/HotspotListItem.tsx
@@ -5,7 +5,7 @@ import { useTranslation } from 'react-i18next'
 import CheckCircle from '@assets/images/check-circle.svg'
 import Attention from '@assets/images/attention.svg'
 import CarotRight from '@assets/images/carot-right.svg'
-import React from 'react'
+import React, { memo } from 'react'
 import TouchableOpacityBox from './BSTouchableOpacityBox'
 import Box from './Box'
 import Text from './Text'
@@ -91,4 +91,4 @@ const HotspotListItem = ({
   )
 }
 
-export default HotspotListItem
+export default memo(HotspotListItem)
--- a/src/features/hotspots/root/HotspotsList.tsx
+++ b/src/features/hotspots/root/HotspotsList.tsx
@@ -1,13 +1,16 @@
-import React, { useCallback, useMemo } from 'react'
+import React, { memo, useCallback, useMemo } from 'react'
 import { BottomSheetSectionList } from '@gorhom/bottom-sheet'
 import { Hotspot } from '@helium/http'
 import Balance, { CurrencyType } from '@helium/currency'
 import { useSelector } from 'react-redux'
+import { useTranslation } from 'react-i18next'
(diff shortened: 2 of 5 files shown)


allenan:
I'm not sure this fully addresses the feedback in #253 but from reading that thread, I'm not sure what the desired UX is exactly. @cokes518 ?

cokes518:
> show "No Hotspots Offline" text for offline filter without offline hotspots

To expand on this, if you apply the offline filter and there is 1 or more Hotspots offline, then we should *only* show those offline hotspots, and none of the online spots.

If you apply the offline filter and there are NO hotspots offline, then we state "No offline hotspots", with a list of online hotspots below. I feel presenting the data this way will prevent people from accidentally selecting the filter then panicking that all their hotspots are gone from their account.

tyler-whitman:
> > show "No Hotspots Offline" text for offline filter without offline hotspots
> 
> To expand on this, if you apply the offline filter and there is 1 or more Hotspots offline, then we should _only_ show those offline hotspots, and none of the online spots.
> 
> If you apply the offline filter and there are NO hotspots offline, then we state "No offline hotspots", with a list of online hotspots below. I feel presenting the data this way will prevent people from accidentally selecting the filter then panicking that all their hotspots are gone from their account.

Ok sounds good. So thats what I was doing except I was not filtering out the online ones in the first case, they were just sorted by offline first. Let me make that change.

allenan (approved):
this works well and would be fine to merge as is. I'm wondering about the consistency of the experience though. It's a bit inconsistent that when I have offline hotspots, the "Offline Hotspots" filter shows just my offline hotspots, but when I don't have any offline hotspots, I see "No offline hotspots" and then all of my online hotspots under "online hotspots". I guess I'd sort of expect to see maybe just the "no offline hotspots" message maybe larger in the center with an emoji for flair. Just an idea though

tyler-whitman:
> this works well and would be fine to merge as is. I'm wondering about the consistency of the experience though. It's a bit inconsistent that when I have offline hotspots, the "Offline Hotspots" filter shows just my offline hotspots, but when I don't have any offline hotspots, I see "No offline hotspots" and then all of my online hotspots under "online hotspots". I guess I'd sort of expect to see maybe just the "no offline hotspots" message maybe larger in the center with an emoji for flair. Just an idea though

Yeah I was thinking this too, but followed what Coco wanted it to show. I think lets merge it and see what people thing and we can change it if needed

PR #277, opened 2021-02-23
Show other nearby hotspots on the map

### LINKED ISSUE #271: Show Nearby Hotspots
On a Hotspot map, plot nearby hotspots visible in view. As the user zooms out, display and query for more hotspots up to a resolution 4. 

This feature will depend on #20 and will be a new entry point for #50.

Will require some designs from @PeterMain in terms on interaction once a user taps on a dot (name, city, reward scale, online/offline, tap to view details, favorite/unfavorite).

image

allenan:
resolves #271 

open questions:
- are we ok with the zoomed out style of network hotspots? should there be a different style lower zoom levels?
- should we exclude offline hotspots at the server level? would reduce the number of items on the map and more accurately represent nearby coverage
- is the zoom/fly effect when toggling witnesses too intense? there's the possibility it disrupts the user's flow when exploring around
- we also need to see how the server stands up to these requests under load, tbd


cokes518:
> are we ok with the zoomed out style of network hotspots? should there be a different style lower zoom levels?

this looks fine in the video. happy with it.
> should we exclude offline hotspots at the server level? would reduce the number of items on the map and more accurately represent nearby coverage

maybe we can color them differently? too many queries? want to stay consistent with explorer that will use red for offline hotspots.

> is the zoom/fly effect when toggling witnesses too intense? there's the possibility it disrupts the user's flow when exploring around

hm, I haven't noticed this to be an issue and my hotspots have far-away witnesses. maybe we can optimize later

> we also need to see how the server stands up to these requests under load, tbd

ok

src/components/Map.tsx
```diff
@@ -66,22 +73,44 @@ const Map = ({
...
   const colors = useColors()
   const { t } = useTranslation()
   const map = useRef<MapboxGL.MapView>(null)
   const camera = useRef<MapboxGL.Camera>(null)
   const [loaded, setLoaded] = useState(false)
   const [userCoords, setUserCoords] = useState({ latitude: 0, longitude: 0 })
   const [centerOffset, setCenterOffset] = useState(0)
+  const styles = useMemo(() => makeStyles(colors), [colors])
+
+  const {
```
matthewcarlreetz, line 89: Seeing as though this is a pretty lightweight object, I'd recommend using lodash's `isEqual` function here
```ts
 const networkHotspots = useSelector(
    (state: RootState) => state.networkHotspots.networkHotspots,
    isEqual,
  )
```

src/components/Map.tsx
```diff
@@ -274,19 +320,29 @@ const Map = ({
...
+            style={styles.markerNetwork}
+          />
+        </MapboxGL.ShapeSource>
         <MapboxGL.ShapeSource
           id="ownedHotspots"
           shape={shapeSources.ownedHotspotFeatures}
           onPress={onShapeSourcePress}
         >
-          <MapboxGL.SymbolLayer id="markerOwned" style={styles.markerOwned} />
+          <MapboxGL.CircleLayer id="markerOwned" style={styles.markerOwned} />
```
tyler-whitman, line 338: Nice much better than using PNGs

matthewcarlreetz (approved):

src/components/Map.tsx
```diff
@@ -66,22 +73,44 @@ const Map = ({
...
+    }
+  }, [loaded, dispatch])
 
   const onRegionDidChange = useCallback(async () => {
     if (onMapMoved) {
       const center = await map.current?.getCenter()
       onMapMoved(center)
     }
-  }, [onMapMoved])
+    if (showNearbyHotspots) {
```
tyler-whitman, line 107: So does this get called whenever you move the map?

src/store/rootReducer.ts
```diff
@@ -17,6 +18,7 @@ const rootReducer = combineReducers({
   hotspotDetails: hotspotDetailsSlice.reducer,
   hotspots: hotspotsSlice.reducer,
   hotspotChecklist: hotspotChecklistSlice.reducer,
+  networkHotspots: networkHotspotsSlice.reducer,
```
tyler-whitman, line 21: Maybe we should cache this reducer using redux-persist or something?

tyler-whitman:
I think we should remove either the settings all together for the network hotspots or at least remove the transfer hotspot option when you go to settings.

cokes518:
Comment out the settings cuz we’ll use it later for favorite a hotspot

________________________________

src/components/Map.tsx
allenan: yeah it'll load whatever is in the visible map bounds which can be quite a few if you zoom out. it seemed to perform ok for me still though. and once it has them it keeps them in temporary memory

src/store/rootReducer.ts
allenan: i was thinking about that, but figured for now it'd be easier to get out of a pickle by having someone force quit the app, in case something bad got stuck in there. we can revisit that if performance becomes an issue either in the app or the api

PR #288, opened 2021-02-23
remove background location permission on Android

tyler-whitman:
It was coming from expo-location https://docs.expo.io/versions/latest/sdk/location/


allenan (approved):
we'll want to test for regressions on bluetooth functionality since that depends on the location permission

tyler-whitman:
Yeah I gave it a quick test and I was able to connect to my hotspot still and sort by close hotspots. We weren’t using background locations anyway so we should be good but doesn’t hurt to test more.

PR #289, opened 2021-02-23
fetch account chart data when new blocks come in

allenan:
Adds chart data to the data that gets fetched on initial load and when new blocks come in. I noticed that the chart data wasn't updating when I was testing pending transactions. This should make it consistent with the timing of the other data we fetch.

### DIFF
--- a/src/components/BarChart/WalletChart.tsx
+++ b/src/components/BarChart/WalletChart.tsx
@@ -1,6 +1,6 @@
 import React, { memo, useCallback, useEffect, useMemo, useState } from 'react'
 import { ActivityIndicator, TouchableWithoutFeedback } from 'react-native'
-import { round } from 'lodash'
+import { isEqual, round } from 'lodash'
 import { useSelector } from 'react-redux'
 import ChartContainer from './ChartContainer'
 import CarotLeft from '../../assets/images/carot-left.svg'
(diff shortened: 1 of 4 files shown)


src/components/BarChart/WalletChart.tsx
```diff
@@ -22,32 +24,33 @@ type Props = {
 const WalletChart = ({ height }: Props) => {
   const dispatch = useAppDispatch()
   const {
-    account: { activityChart },
+    account: { activityChart, activityChartRange },
     activity: { filter },
   } = useSelector((state: RootState) => state)
```
matthewcarlreetz, line 29: I think this change would eliminate renders happening every time the store is updated, but no actual change happens

``` javascript
  const {
    account: { activityChart, activityChartRange },
    activity: { filter },
  } = useSelector(
    (state: RootState) => state,
    (prev, next) => {
      const activityChartEqual = isEqual(
        prev.account.activityChart,
        next.account.activityChart,
      )
      const rangeEqual =
        prev.account.activityChartRange === next.account.activityChartRange
      const filterEqual = prev.activity.filter === next.activity.filter

      return activityChartEqual && rangeEqual && filterEqual
    },
  )
```

matthewcarlreetz (approved):
Looks good, just one suggestion.

src/components/BarChart/WalletChart.tsx
allenan: ok thanks will play around with that and see if it improves it

src/utils/logger.ts
```diff
@@ -14,6 +14,9 @@ export const setUser = (userAddress: string) => {
 }
 
 export const error = (e: unknown) => {
+  if (__DEV__) {
```
matthewcarlreetz, line 17: yes!

matthewcarlreetz (approved):

Issue #75, opened 2021-01-05
Favorite Hotspot: Update Wi-Fi


cokes518:
Once a user has paired over bluetooth with a favorited Hotspot, allow them to change the Wi-Fi credentials.

cokes518:
can already do this. closing

Issue #74, opened 2021-01-05
Favorite Hotspot: Run Diagnostic


cokes518:
Allow a user, once paired, to run a diagnostic report of a favorited Hotspot.

Once the report is generated, users will be able to copy and email contents.

Issue #73, opened 2021-01-05
Favorite Hotspot: Pair over Bluetooth


cokes518:
Allow a user to pair with a favorited Hotspot over bluetooth to unlock further settings.

Issue #72, opened 2021-01-05
Favorite Hotspot: View Details


cokes518:
Followed Hotspots will show the same data as owned Hotspots.

- view online/offline status
- view earnings
- checklist
- view witnesses, including map
- view challenges passed
- will display "Owned by " instead of "Owned by you"

PR #295, opened 2021-02-23
Hosted hotspot connect message update.

cvolkernick:
Small change to the message shown to a user setting up a hotspot they do not personally own. This flow is currently the only way (outside of Kent's unofficial hotspot utility) hotspot hosts can connect their hotspots to Wi-Fi without owner involvement, and the dialog is particularly disorienting to less technically inclined users. Preferably the dialog here would also be able to be closed not backed out of, as the latter intuitively indicates that the network connection did not succeed.

### DIFF
--- a/src/locales/en.ts
+++ b/src/locales/en.ts
@@ -313,7 +313,7 @@ export default {
       already_added:
         'You already added this hotspot to your wallet. Continue to the next screen to assert its location.',
       not_owned:
-        'You do not own this hotspot and cannot add it to your wallet.',
+        'It looks like you’re adding a hosted hotspot; you can connect to Wi-Fi but it will not show in your app.',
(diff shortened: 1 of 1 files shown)


cokes518:
I'm not sure this is the right generalization to make (assume all Hotspots are hosted). 

In the coming future, the new app will allow hosts to favorite or follow a hotspot they do not own, and certain settings like update wi-fi will be exposed without having to go through the onboarding flow as a workaround.

You can see the list of support issues here: #50 , #71 , #72, #73, #74, #75

allenan:
This will be addressed more comprehensively in upcoming features

cvolkernick:
https://discord.com/channels/404106811252408320/796071060172308480/813901947517272074 thought so....

PR #304, opened 2021-02-25
selectable address component

### LINKED ISSUE #164: In my wallet payment transactions you cannot copy wallet addresses
On the previous version the from and to addresses could be copied to allow easy sending of HNT especially as the addresses are truncated in the viewimage

tyler-whitman:
On click of an address in the app you will now get a bottom sheet with actions to copy it or view on explorer. On long press it will automatically copy and send a test message confirming the address has been copied.

Screen Shot 2021-02-24 at 6 50 01 PM

closes #164


src/components/Address.tsx
```diff
@@ -0,0 +1,159 @@
...
+  ]
+
+  const onPress = () => {
+    if (clickToCopy) {
+      copyAddress()
+    } else {
+      showActionSheetWithOptions(
+        {
+          options: opts.map(({ label }) => label),
+          destructiveButtonIndex: opts.length - 1,
```
allenan, line 103: might want to change this to `cancelButtonIndex` to get the following behavior:
image

tyler-whitman: Ok I copied the one I have from the settings and it shows it in red. Which one do we prefer?

tyler-whitman: I guess cancel makes more sense?

allenan: yeah let's go with the cancel one. seems more idiomatic since it's not a destructive action

tyler-whitman: Sounds good

allenan (approved):
works great 👍

PR #308, opened 2021-02-25
setting: update text for new 3rd party options

shawaj:
Update the text to refer to all 3rd party devices.

Change-type: patch
Signed-off-by: Aaron Shaw <shawaj@gmail.com>

### DIFF
--- a/src/locales/en.ts
+++ b/src/locales/en.ts
@@ -591,7 +591,7 @@ export default {
     pairing: {
       title: 'Hotspot Pairing Required',
       subtitle:
-        'Press the button on your Helium Hotspot, or power-cycle your RAK Hotspot Miner.',
+        'Press the button or power-cycle your Helium or 3rd party Hotspot.',
       scan: 'Scan for my Hotspot',
     },
     transfer: {


matthewcarlreetz (commented):
Thanks. Good catch on this one. 

@cokes518 We'll need to update the translations as well.

src/locales/en.ts
```diff
@@ -591,7 +591,7 @@ export default {
     pairing: {
       title: 'Hotspot Pairing Required',
       subtitle:
-        'Press the button on your Helium Hotspot, or power-cycle your RAK Hotspot Miner.',
+        'Press the button or power-cycle your Helium or 3rd party Hotspot.',
```
cokes518, line 594: this implies that you need to power cycle the helium hotspot.

Recommended change ->
"Place your Hotspot in Bluetooth pairing mode."

owenselles: agree with this suggested change

tyler-whitman: @shawaj do you want to update this? Or want us to do it?

tyler-whitman:
Closing as I made the change in https://github.com/helium/hotspot-app/pull/368

PR #313, opened 2021-02-26
pin for payment

### LINKED ISSUE #302: PIN required for Payment not implemented
The app doesn't respect "require pin for payment"... we should fix that.

tyler-whitman:
closes #302 

- Fixes require pin for payments
- Require pin when navigating to send screen from wallet
- Require pin when navigating to send from payment QR 
- Updates the Android toggle colors in settings to match iOS

Screen Shot 2021-02-25 at 5 05 22 PM

### DIFF
--- a/src/features/moreTab/more/MoreListItem.tsx
+++ b/src/features/moreTab/more/MoreListItem.tsx
@@ -1,4 +1,4 @@
-import React, { memo } from 'react'
+import React, { memo, useMemo } from 'react'
 import { Linking, Switch } from 'react-native'
 import RNPickerSelect, { Item } from 'react-native-picker-select'
 import Text from '../../../components/Text'
@@ -51,6 +51,11 @@ const MoreListItem = ({
     }
   }
 
+  const trackColor = useMemo(
+    () => ({ false: colors.purple400, true: colors.purpleMain }),
+    [colors],
+  )
+
   return (
     <TouchableOpacityBox
       flexDirection="row"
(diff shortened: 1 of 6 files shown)


src/features/wallet/send/SendNavigator.tsx
```diff
@@ -1,15 +1,27 @@
...
 const SendStack = createStackNavigator()
 
-const SendNavigator = () => {
+type Route = RouteProp<RootStackParamList, 'Send'>
+
+type Props = {
+  route: Route
+}
+const SendNavigator = ({ route }: Props) => {
+  const scanResult = route?.params?.scanResult
```
matthewcarlreetz, line 17: I think this could be written like this to avoid the inline object

```ts
  const { scanResult } = route?.params || {}
  return (
    <SendStack.Navigator headerMode="none">
      <SendStack.Screen
        name="Send"
        component={SendScreen}
        initialParams={scanResult}
      />
```

matthewcarlreetz (changes requested):

src/features/moreTab/more/MoreListItem.tsx
```diff
@@ -78,7 +78,8 @@ const MoreListItem = ({
         <Switch
           value={value as boolean}
           onValueChange={onToggle}
-          trackColor={{ false: colors.purpleMain, true: colors.purpleMain }}
+          trackColor={{ false: '#303351', true: colors.purpleMain }}
```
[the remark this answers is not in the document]
tyler-whitman: Updated

src/features/wallet/send/SendNavigator.tsx
tyler-whitman: Udpated

matthewcarlreetz (approved):

PR #322, opened 2021-02-26
remove noisy sentry logs and cleanup initial load

### LINKED ISSUE #320: Prevent common errors from sending to Sentry
We have a few super common errors being reported in Sentry that aren't actually problematic.

- https://sentry.io/organizations/helium-su/issues/2228550073/?project=5584284&query=is%3Aunresolved
- https://sentry.io/organizations/helium-su/issues/2227817876/?project=5584284&query=is%3Aunresolved
- https://sentry.io/organizations/helium-su/issues/2227779560/?project=5584284&query=is%3Aunresolved
- https://sentry.io/organizations/helium-su/issues/2188032202/?project=5584284&query=is%3Aunresolved

We should either just ignore these types of errors, or figure out the root cause behind why they are so prevalent and fix that.

tyler-whitman:
closes #320

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,13 +1,13 @@
 import 'react-native-gesture-handler'
-import React, { useEffect, useCallback } from 'react'
+import React, { useCallback, useEffect } from 'react'
 import { SafeAreaProvider } from 'react-native-safe-area-context'
 import {
-  Platform,
-  StatusBar,
   AppState,
   AppStateStatus,
-  UIManager,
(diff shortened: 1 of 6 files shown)


matthewcarlreetz (approved):
Looks good. Do you think we should make the logger changes for the other clients as well?

tyler-whitman:
> Looks good. Do you think we should make the logger changes for the other clients as well?

Good point let me update

Issue #307, opened 2021-02-25
Unable to drop a pin or enter an address on assert page


cokes518:
A few users have reported that they were unable to enter an address in the address field or drop a pin by panning around on the map. This is during the add hotspot flow.

This is happening on iOS and android, wallet address and physical address used in the discord message: https://discord.com/channels/404106811252408320/733699201887764511/814547381373894707

timtamothee:
This issue may be only for specific Hotspots. I found I'm not able to assert the Hotspot in any location because the wheel doesn't stop spinning to load. The Hotspot ID is 4C:E2.

allenan:
@timtamothee which wheel is this? would you be able to dm me a video showing this behavior?

cokes518:
Update: Hotspots not on the onboarding server will encounter this issue. 

We'll update the error message to be more meaningful and communicate next steps.

"Unable to find Hotspot in the Onboarding Server. Please contact the Hotspot manufacturer for next steps."

PR #324, opened 2021-02-26
add onboarding error screen to hotspot setup flow

allenan:
at one point we were throwing an alert error when onboarding record couldn't be found. The logic to trigger this alert changed and it was no longer getting triggered, leading the user to get further along in the setup than they should have. This adds a proper screen for onboarding error, and navigates to it when there is no onboarding record.

resolves: #307

### DIFF
--- a/src/features/hotspots/setup/HotspotSetupConfirmLocationScreen.tsx
+++ b/src/features/hotspots/setup/HotspotSetupConfirmLocationScreen.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useCallback, useEffect } from 'react'
 import { ActivityIndicator } from 'react-native'
 import { useTranslation } from 'react-i18next'
 import { RouteProp, useNavigation, useRoute } from '@react-navigation/native'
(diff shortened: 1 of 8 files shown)


src/features/hotspots/setup/HotspotSetupConfirmLocationScreen.tsx
```diff
@@ -30,15 +31,22 @@ const HotspotSetupConfirmLocationScreen = () => {
   const {
     account: { account },
   } = useSelector((state: RootState) => state)
-  const { loading, result } = useAsync(loadLocationFeeData, [])
+  const { loading, result, error } = useAsync(loadLocationFeeData, [])
```
allenan, line 34: this is where Tim was getting stuck, so I added a case to check the error, but it actually should error out sooner than this screen

src/utils/useHotspot.ts
```diff
@@ -184,6 +185,11 @@ const useHotspot = () => {
     }
 
     const response = await dispatch(fetchHotspotDetails(details))
+    if (
+      !(response.payload as AllHotspotDetails).onboardingRecord?.onboardingKey
```
allenan, line 189: we now make sure we have an onboarding record and return false if not, triggering the navigation to the error screen

tyler-whitman (approved):

PR #333, opened 2021-03-02
enable hotspot onboarding without location assert

### LINKED ISSUE #305: Missing skip location assert during initial hotspot setup
The new 3.0 app does not have the ability to skip location assert during initial setup. Hotspots can't be added to account now without a location set. Ver 2 was able to skip location for the hotspot to be added to the account.

tyler-whitman:
closes #305 

Enables a way to skip location assert when onboarding a hotspot. The new flow is shown below

Screen Shot 2021-03-01 at 6 35 36 PM


src/features/hotspots/setup/HotspotSetupLocationInfoScreen.tsx
```diff
@@ -22,6 +23,10 @@ const HotspotSetupLocationInfoScreen = () => {
     }
   }
 
+  const skipLocationAssert = async () => {
```
matthewcarlreetz, line 26: probably doesn't matter, but could wrap this in a `useCallback`

src/locales/en.ts
```diff
@@ -396,6 +396,11 @@ export default {
       alertMessage:
         'Request to servers have timed out and we cannot add your Hotspot at this time.\n\nPlease contact support@helium.com and note MAC address %{mac}.',
     },
+    skip_location: {
```
matthewcarlreetz, line 399: We should add these to the other language files as well. cc: @cokes518

matthewcarlreetz:
Looks good. My ble changes have caused some conflicts for you. Sorry :(

src/locales/en.ts
cokes518: they'll auto populate when we send the en.json out for translation.

src/features/hotspots/setup/HotspotSetupLocationInfoScreen.tsx
tyler-whitman: Do we need it for things that dont use any state or anything? Would rather keep it simple if it doesn't really affect anything

matthewcarlreetz (approved):

PR #336, opened 2021-03-02
Show better errors when bluetooth ops fail.

### LINKED ISSUE #323: Better error message when bluetooth disconnects
Right now our default error message for anything is "Something went wrong." which is not helpful at all.

In this specific use case, if bluetooth disconnects (sentry logs say `BleError: BluetoothLE is powered off`), we should show "Bluetooth Disconnected. Please pair again."

This was found while testing forget wi-fi bug fix.

### LINKED ISSUE #334: Assert Location - RangeError
Ryan from Nebra reported an issue with assertion while onboarding a hotspot. It seems to be a more general bug, but is specific to iOS and has affected ~75 users.

[Sentry Link](https://sentry.io/share/issue/38f43cfa386d4956b9cdf16db7acd4c0/)

A BLE read is performed after writing to the add gateway characteristic, that response value is used to construct a gateway transaction. 
`const addGatewayTxn = AddGatewayV1.fromString(partialTxnBin)`
This line is throwing a RangeError exception. 
> RangeError: index out of range: 3 + 8 > 4

IMG_0050

matthewcarlreetz:
closes #323 
closes #334

### DIFF
--- a/src/utils/bluetooth/useBluetooth.tsx
+++ b/src/utils/bluetooth/useBluetooth.tsx
@@ -6,7 +6,6 @@ import {
   Device,
   LogLevel,
 } from 'react-native-ble-plx'
-import { decode } from 'base-64'
 import sleep from '../sleep'
 import {
   FirmwareCharacteristic,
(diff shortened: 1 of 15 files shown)


src/utils/bluetooth/useBluetooth.tsx
```diff
@@ -124,44 +124,46 @@ const useBluetooth = () => {
...
-        `Successfully read Characteristic: ${characteristic.uuid} for service: ${characteristic.serviceUUID}`,
-      )
+      if (errorMessage) {
+        const error = new Error(errorMessage)
+        Logger.error(error)
+        throw error
+      }
 
-      return value
+      return charWithValue
```
tyler-whitman, line 162: Should this be `parsedValue`?

tyler-whitman: Oh wait never mind, this is correct

tyler-whitman (approved):
Looks good, lets hope this fixes that crash

PR #346, opened 2021-03-03
fix more onboarding issues

tyler-whitman:


src/utils/useHotspot.ts
```diff
@@ -325,24 +328,13 @@ const useHotspot = () => {
     return compareVersions.compare(deviceFirmwareVersion, minVersion, '>=')
   }
 
-  const updateHotspotStatus = async () => {
-    const { address } = connectedHotspotDetails
-    if (!address) return
-
+  const updateHotspotStatus = async (hotspot?: Hotspot) => {
```
tyler-whitman, line 331: I couldnt find anywhere this was actually getting called, and the status was never set. I decided to just call it as part of the initial details fetch.

tyler-whitman: Its here if you missed it https://github.com/helium/hotspot-app/pull/346/files?diff=unified&w=1#diff-812d548c99bb785e02911eb20cac11852b4c9018115489a15f5b8c91f9905067R200

matthewcarlreetz (approved):

PR #349, opened 2021-03-04
Feature/48 hnt to currency

### LINKED ISSUE #48: Represent USD value of a wallet's HNT in-app.
**User story:** One of the most common actions I take is to check the overall balance of my HNT in the wallet app and copy it to Blockfolio to get a sense of my 'portfolio' value. It would save me time and be more convenient to reference the wallet value only in the hotspot app. This would help the broader app user base as many users are unlikely to know that Blockfolio is available for this type of calculation, and furthermore there are a set of users that are unlikely to know where to find the market value of HNT at all.

**Proposed solution:** It would be nice to be able to toggle HNT balance in the app, backed by something like CMC, Oracle Price, or another source. I feel the source should be represented to the user in some manner so they can continue their exploration outside of the app, reinforcing that the USD value isn't just made up.
Depending on the layout of the new app UI, the USD value could be included as parenthetical to the HNT balance, noted in some sort of tooltip, or toggled on a tap like the price representations in the Robinhood app. 

As an extension of this though, I would be careful around making the dollar value the most prominent aspect of the balance. The USD value of HNT fluctuates wildly, as does the rest of the crypto market. I think significant drops are likely to create a bit of anxiety and more "Why are my earnings down".

matthewcarlreetz:
closes #48 

Adds a setting on the more tab to toggle currency/hnt
Adds a touch interaction to hnt values to toggle currency display
Uses the phone's locale to determine currency type


src/components/HotspotListItem.tsx
```diff
@@ -25,69 +24,74 @@ const HotspotListItem = ({
   showCarot = false,
 }: HotspotListItemProps) => {
   const { t } = useTranslation()
-
+  const { toggleConvertHntToCurrency } = useCurrency()
+  const handlePress = useCallback(() => onPress?.(hotspot), [hotspot, onPress])
```
tyler-whitman, line 28: Can we add a debounce to this? Seems to take some time so spamming it might be bad

src/components/HotspotListItem.tsx
```diff
@@ -25,69 +24,74 @@ const HotspotListItem = ({
...
+              </Text>
+            </Box>
             <Text
-              variant="body2Medium"
-              color="offblack"
-              paddingStart="s"
-              ellipsizeMode="tail"
-              numberOfLines={2}
-              maxWidth={210}
+              onPress={toggleConvertHntToCurrency}
```
tyler-whitman, line 72: Can we add a debounce to this

matthewcarlreetz: yeah, good call

tyler-whitman: I guess you added the ability to tap from multiple places which is awesome. Maybe the debounce can be in a lower level so we dont have to add it everywhere

matthewcarlreetz: yeah, I think I'll debounce the `toggleConvertHntToCurrency` function.

src/features/wallet/send/SendView.tsx
```diff
@@ -143,8 +144,9 @@ const SendView = ({ scanResult, sendType, hotspot, isSeller }: Props) => {
   // compute equivalent dc amount for burn txns
   useAsync(async () => {
     if (type === 'dc_burn') {
-      const balanceDc = await networkTokensToDataCredits(balanceAmount)
-      // TODO might need to round up in DC conversion in he-js
+      const balanceDc = networkTokensToDataCredits(balanceAmount)
+      if (!balanceDc) return
```
tyler-whitman, line 148: Just wondering why you added this in? Existing bug?

src/utils/useCurrency.ts
```diff
@@ -0,0 +1,142 @@
...
+    split?: true,
+    maxDecimalPlaces?: number,
+  ) => { integerPart: string; decimalPart: string }
+  const hntBalanceToDisplayVal = useCallback(
+    (
+      balance: Balance<NetworkTokens>,
+      split?: boolean,
+      maxDecimalPlaces = 2,
+    ) => {
+      const multiplier = currentPrices?.[currencyType.toLowerCase()] || 0
```
tyler-whitman, line 93: instead of `|| 0` should we do `currentPrices?.['usd']`?

tyler-whitman: Oh I see the `|| 0` would just keep it in HNT. Actually maybe thats better than USD

tyler-whitman (approved):
Looks great!

PR #356, opened 2021-03-05
Add alternative characteristic for Firmware / Software Version

### LINKED ISSUE #354: Firmware Characteristic Service Not Suitable To Be Used (?)
Today in preparation of getting our hotspots ready to be programmed I done some last steps of updating our base OS software to the latest version which brought a breaking bug for us. However the issue could manifest in the future as others update as well.

While we are using our own BTLE implementation, this same issue is re-producable with the gateway_config tool provided here on github due to the nature of the issue.

Currently for the firmware version the GATT characteristic `2A26` is used under the Device Information Service `180A`, however in version 5.55 of Bluez and above a service is provided by bluetoothd on 180A which causes a conflict as the app searches for the 2A26 characteristic under the service provided by bluez and fails.

Upon contacting the maintainers of Bluez so far the response is that it doesn't seem to be recommended that a user's app provides this service:

```
> However as our software creates a service on 0x180A it causes a
> conflict, as the application we are making our software with reads
> from the first service available which is the one generated by BlueZ.

You should probably stop doing that and just use main.conf to set the
values you want to be published, we should actually blacklist DIS to
be registered by applications since that is defined by the platform
having the application to override may actually cause problems.
```

The above also hints that they may block the service from being registered at all in the future.

In a brief discussion with @madninja one suggestion was to pin the bluez version to an earlier one (5.53 was working on our older software, however as the Base OS that provides Blue Z isn't under our control with the OTA mangement platform we're using, we're currently unable to do this). 

However I could see the above becoming an issue over time as pinning to an older version of software isn't going to be possible in the long run.

The best solution that I can think of is to just move the firmware request characteristic to the custom service everything else is under as this would then future proof the code, however it does mean existing hotspots would need updating with this change.  However the code change in gateway_config does look fairly trivial to do. 

Or possibly if it fails the check with the characteristic in the device info service, that it then requests it from a second one to then provide good backwards compatibility.

ryanteck:
Following up from the issue I created at https://github.com/helium/hotspot-app/issues/354 

I implemented a solution for an alternative characteristic (called Software Version) that is called at the time it checks for firmware.

If the original firmware characteristic is returned it uses that. If that isn't returned it uses the new characteristic as a fallback. And if neither are returned it fails like it should as before.

By implementing it in this way, hotspots that are able to use the old method still work as intended (which I've tested).

And as in our case we can't no longer use this service to provide the characteristic, the new one is in the same custom service which is working perfectly fine.

closes #354

### DIFF
--- a/src/utils/bluetooth/bluetoothDataParser.ts
+++ b/src/utils/bluetooth/bluetoothDataParser.ts
@@ -52,7 +52,8 @@ export function parseChar(
     | typeof HotspotCharacteristic.WIFI_REMOVE
     | typeof HotspotCharacteristic.WIFI_CONNECT_UUID
     | typeof HotspotCharacteristic.ADD_GATEWAY_UUID
-    | typeof FirmwareCharacteristic.FIRMWAREVERSION_UUID,
+    | typeof FirmwareCharacteristic.FIRMWAREVERSION_UUID
+    | typeof HotspotCharacteristic.SOFTWARE_VERSION_UUID,
 ): string
 export function parseChar(
   characteristicValue: string,
--- a/src/utils/bluetooth/bluetoothTypes.ts
+++ b/src/utils/bluetooth/bluetoothTypes.ts
@@ -18,4 +18,5 @@ export enum HotspotCharacteristic {
   ASSERT_LOC_UUID = 'd435f5de-01a4-4e7d-84ba-dfd347f60275',
   DIAGNOSTIC_UUID = 'b833d34f-d871-422c-bf9e-8e6ec117d57e',
   ETHERNET_ONLINE_UUID = 'e5866bd6-0288-4476-98ca-ef7da6b4d289',
+  SOFTWARE_VERSION_UUID = 'c0b64050-697d-463a-a33f-70c4825731f8',
 }
--- a/src/utils/useHotspot.ts
+++ b/src/utils/useHotspot.ts
@@ -312,9 +312,23 @@ const useHotspot = () => {
       connectedHotspot.current,
       Service.FIRMWARESERVICE_UUID,
     )
-    if (!charVal) return false
 
-    const deviceFirmwareVersion = parseChar(charVal, characteristic)
+    const characteristicAlt = HotspotCharacteristic.SOFTWARE_VERSION_UUID
+    const charValAlt = await findAndReadCharacteristic(
+      characteristicAlt,
+      connectedHotspot.current,
+      Service.MAIN_UUID,
+    )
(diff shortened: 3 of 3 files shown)


tyler-whitman (approved):
Change looks good to me, so seems we should be updating this in our bluetooth code as well then?

ryanteck:
By doing it this way @tyler-whitman it means that existing code doesn't need updating as it first requests for the original value, then in cases where it fails it can fall back to the new method I proposed.

So essentially the current code on deployed hotspots and unless then the BlueZ on them gets updated doesn't need to be updated. However if other manufacturers are running newer versions of the bluetooth stack like we most likely will be this resolves it.

tyler-whitman:
> By doing it this way @tyler-whitman it means that existing code doesn't need updating as it first requests for the original value, then in cases where it fails it can fall back to the new method I proposed.
> 
> So essentially the current code on deployed hotspots and unless then the BlueZ on them gets updated doesn't need to be updated. However if other manufacturers are running newer versions of the bluetooth stack like we most likely will be this resolves it.

Yeah I realize that, I just mean we should probably update as we are using an approach that is not recommended and may cause bugs.

ryanteck:
> Yeah I realize that, I just mean we should probably update as we are using an approach that is not recommended and may cause bugs.

Indeed it would possibly make sense to migrate all to the new characteristic but it's still slightly unclear which is the "proper" way to do it. 

I think from discussions I've had with @madninja that the current way should be suitable as per say the BTLE spec, but it's then unclear why the Linux Bluetooth devs are saying it's unsuitable. I'm going to try and pop on IRC to see if I can get a quicker response from them as the suggested solution from them didn't work at all. 

As well with the proposed solutions from them would likely require files to be updated if required on the base os itself rather than say just the software that runs the Bluetooth communication which the above proposed solution also fixes.

Currently so far without pinning Blue Z to an older version (something we're working with our OTA software solution to see if we can do as they bumped it to 5.55 which caused this to break) I couldn't figure out another solution.

ryanteck:
So it looks like BlueZ have now patched the issue so that in the future it will be fixed and work like it used to.(https://github.com/bluez/bluez/issues/101)

However I'm still trying to see if we can get BlueZ pinned to an older version on the OS we're using which is taking longer than I would have hoped. So ideally if possible if this could be accepted as the workaround doesn't affect existing devices from the best I can test.

ryanteck:
I believe we've now been able to pin to and older version of Blue Z and as it should be patched in future versions of BlueZ believe this can be closed now.

PR #360, opened 2021-03-06
enable accounts without hotspots to view hotspots around them

### LINKED ISSUE #184: Use fitBound + padding for witness zoom calculation
We should investigate using the Camera.fitBounds function with a bottom padding offset for the witness zoom calculation. Right now the map centers on the screen, but the card is covering most of it

https://github.com/react-native-mapbox-gl/maps/blob/master/docs/Camera.md#fitboundsnortheastcoordinates-southwestcoordinates-padding-animationduration

### LINKED ISSUE #314: App shows Hotspot (that I don't own) when tapping "My Hotspots" card
This is likely an edge case.

iOS, TF139
Steps to reproduce:
1. Go to any Hotspot you own
2. tap and explore a Hotspot (that is not yours) on the hotspot map
3. go back to hotspot tab
4. tap "My Hotspot" card
5. see the hotspot from step 2.

Expected results:
See my Hotspots.

### LINKED ISSUE #337: Show Hotspots around you even if you don't own any
As a way for new users to explore their geographical area, we should display a map of hotspots around their device location so they can:
1. better understand neighboring hotspots and their stats 
2. determine if there are any close by
3. or as a host, find and interact with the hotspot the have in their home

Use case 3 will serve as the initial entry point to follow hotspot (placeholder until search is implemented). By selecting a Hotspot from the map, they can enter bluetooth pairing mode to set up wi-fi or run diagnostics.

image

tyler-whitman:
closes #337 
closes #314 
closes #184 

Allows users without hotspots in their to view hotspots around them. This will be more beneficial when we also add the ability to favorite hotspot.

This also makes some changes to the map so that you retain your selected hotspot if you go back and forth. So if I select one and go back to the list and then swipe the list down the map no longer resets. It will reset when you hit "My Hotspots", but I did that by design and we could also keep the previous state if we wanted instead.

Currently the flow works as so
1. If you open the app and have not enabled location permission you see the current hotspot empty screen (no bottom sheet or map)
2. When you enable location you jump into the map at your location. If you permanently enable location you will now start on the bottom sheet and have the ability to see the map.
3. If users only enable permission one time then they will continue to see the old static hotspots page (no bottom sheet or map)

https://user-images.githubusercontent.com/6345962/110196774-a8f7af80-7dfb-11eb-93c2-a16c535ddc4b.mov

### DIFF
--- a/src/components/BackButton.tsx
+++ b/src/components/BackButton.tsx
@@ -5,7 +5,7 @@ import { useTranslation } from 'react-i18next'
 import Text from './Text'
 import { Colors, Spacing, Theme } from '../theme/theme'
 import BackArrow from '../assets/images/backArrow.svg'
-import TouchableOpacityBox from './TouchableOpacityBox'
+import { DebouncedTouchableOpacityBox } from './TouchableOpacityBox'
 import { useColors } from '../theme/themeHooks'
 
 type Props = BoxProps<Theme> & {
@@ -27,7 +27,7 @@ const BackButton = ({
   const colors = useColors()
 
   return (
-    <TouchableOpacityBox
+    <DebouncedTouchableOpacityBox
       onPress={onPress}
       alignSelf="flex-start"
       paddingVertical="s"
@@ -46,7 +46,7 @@ const BackButton = ({
       >
         {t('back')}
       </Text>
-    </TouchableOpacityBox>
+    </DebouncedTouchableOpacityBox>
   )
 }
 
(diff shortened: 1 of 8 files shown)


matthewcarlreetz:
I think I found a little bug. 
1. Select a hotspot
2. Press the witnesses button
3. Select another hotspot
4. Witnesses are not updated for the new hotspot

matthewcarlreetz (approved):
💯

PR #362, opened 2021-03-08
Use native currency formatting

### LINKED ISSUE #359: HNT -> USD  does not work on android

matthewcarlreetz:
Closes #359 

The Hermes js engine doesn't include `Intl`. So, decided to use native currency formatting which forced conversion to need to happen asynchronously.


tyler-whitman (approved):
Working well for me now on both iOS and Android

PR #371, opened 2021-03-10
remove legacy hotspot type

tyler-whitman:
Removes the old hotspot type that was calculated from onboarding address formatting which no longer works.

Replaces it with the onboarding maker name. In the case there is no maker name (DYI or something goes wrong) it will show 'unknown' in diagnostics.


matthewcarlreetz (approved):
Looks good. Seems to work well.

PR #372, opened 2021-03-10
update location permission string and fix state not clearing on logout

### LINKED ISSUE #370: Logout of account doesn't clear state + update location permissions
While testing "view hotspots around me" (as an account with no hotspots), it remembered the last Hotspot I viewed in a previous account so it didn't really show hotspots around my device.

Update this popup to include "view hotspots around me". 
"Allow "Helium" to use your location?"
"Allow Helium to use your location to add your Helium Hotspots to the network and explore Hotspots around you."

tyler-whitman:
closes #370 

I think state of the map / some of the components remains on logout and log back in right away since they are in the same stack, components may not unmount. 

Coco ran into this problem but I couldn't get it to happen on my device. I moved each stack to its own navigator to see if maybe that helps.

### DIFF
--- a/ios/Helium/Info.plist
+++ b/ios/Helium/Info.plist
@@ -38,7 +38,7 @@
 		</dict>
 	</dict>
 	<key>NSLocationWhenInUseUsageDescription</key>
-	<string>Allow Helium to use your location to add your Helium Hotspots to the network</string>
+	<string>Allow Helium to use your location to add your Helium Hotspots to the network and explore Hotspots around you</string>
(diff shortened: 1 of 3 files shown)


matthewcarlreetz (approved):
Seems to work well still. If it doesn't fix the problem, we can always revert back.

tyler-whitman:
> Seems to work well still. If it doesn't fix the problem, we can always revert back.

Yeah, I think when screens are in the same stack components can remain mounted so this was my best guess as to why some state might be staying around.

PR #373, opened 2021-03-10
Bugfix/365 ble android

### LINKED ISSUE #365: Fix "Operation was cancelled" error
On Android after canceling the ble WiFi connect subscription this error is thrown. It's a non-issue, but it's clogging up the logs in Sentry. Need to find a way to ignore this error.

I thought I had it handled on lines 255-269 of
`src/utils/useHotspot.ts`, but apparently, it's not working.

matthewcarlreetz:
closes #365 

Removes the need to scan twice on android when ble is off
Removes error log when wifi subscription is removed on android

### DIFF
--- a/src/utils/bluetooth/useBluetooth.tsx
+++ b/src/utils/bluetooth/useBluetooth.tsx
@@ -35,9 +35,7 @@ const useBluetooth = () => {
 
   const getState = async () => getBleManager().state()
 
-  const enable = async () => {
-    getBleManager().enable()
-  }
+  const enable = async () => getBleManager().enable()
 
   const connect = async (
     hotspotDevice: Device,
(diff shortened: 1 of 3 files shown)


src/utils/bluetooth/useBluetooth.tsx
```diff
@@ -35,9 +35,7 @@ const useBluetooth = () => {
 
   const getState = async () => getBleManager().state()
 
-  const enable = async () => {
-    getBleManager().enable()
-  }
+  const enable = async () => getBleManager().enable()
```
matthewcarlreetz, line 38: this is being awaited, but the function wasn't returning the promise 🤦

tyler-whitman: Ah good catch

tyler-whitman (approved):

PR #374, opened 2021-03-10
Feature/218 custom picker

### LINKED ISSUE #218: Custom picker component
Currently our picker element mimics the iOS scroll wheel picker:
image

This is a little annoying to use because it requires multiple interactions to select a new option: open picker, scroll to the option you want (can only see a few at a time), press done or outside the picker to close.

We could create a custom sheet element for these instances. For example, here's Reddit's picker when choosing a new comment sort order:
image

This only requires 2 interactions to use and you can see all your options at a glance

matthewcarlreetz:
Closes #218 

Used react-native modal and reanimated to create an "ActionSheet" component that replaces all instances of the picker. Initially was going to use `BottomSheetModal`, but ran into lots of collisions with our existing bottom sheets on Android.

Also, fixed some bugs with the more screen along the way. Namely, if you hadn't explicitly set the picker values, they were rendering blank.


tyler-whitman (approved):
Looks great and working well!

PR #378, opened 2021-03-11
dc burn and oracle fixes

### LINKED ISSUE #375: fetch oracle price when a user presses `Send` on wallet page
We currently fetch the oracle price on app start and assume the price is relatively stable throughout the app session. Severe fluctuations in price can affect transaction fee calculations and can cause `Send Max` transactions to fail.

Instead of fetching price on app start, fetch the price again when a user taps the Send button on wallet screen.

### LINKED ISSUE #377: Burning small amount into DCs results amount to be converted to integer
I am trying to burn a small amount of HNT into DC. It seems, that currently the amount is always convert into Int. So for example: if I try to burn 0.00250 HNT, that is automatically converted to 250 HNT.

tyler-whitman:
- Fixes DC burn from QR when using number formats other than United States
- Fixes DC amount rounding issues so that it is always a whole number
- Locks burn data, you can no longer change what is in the QR and would have to generate a new one
- Updates oracle prices when navigating to send for fee / DC calculation

closes #377 
closes #375


allenan (approved):

dpetrov:
thanks for fixing so quickly.

PR #401, opened 2021-03-17
wifi scan again

### LINKED ISSUE #388: Rescan for wifi networks
When connecting to wifi currently the app loads the list and displays it, if the hotspot is still booting or no networks are found on the first scan we just get an empty page.

We should let users know the list may be empty because the hotspot is still booting (common in the case you need a reboot to get to pairing mode) and allow them to rescan for wifi networks without having to reconnect.

Currently on a RAK Hotspot Miner it usually takes about 2-3 minutes for the hotspot to finish boot and display the wifi networks.

tyler-whitman:
closes #388 
Adds the ability to re-scan wifi networks when onboarding and in hotspot settings.

 
Screen Shot 2021-03-17 at 11 45 38 AM
Screen Shot 2021-03-17 at 12 00 27 PM


matthewcarlreetz (approved):
💯

PR #405, opened 2021-03-19
transfer hotspot stale activity check

### LINKED ISSUE #381: do not let transfer proceed if it fails staleness check
Current transfer Hotspots warn against stale hotspots but does not actually *prevent* transfers of stale hotspots. This is a poor user experience as transfers often stay pending or fail silently without a reason.

We should enforce the staleness check that the blockchain uses (chain var `transfer_hotspot_stale_poc_blocks`) that looks for poc activity in the last 1200 blocks. If true, allow the transfer to proceed. If false, do not let buyer press the transfer button and show the reason why: "Hotspot has not had Beacon or Witness activity in the last 1200 blocks."

image

tyler-whitman:
closes #381 

check hotspot for POC activity in the last 1200 blocks (loaded from the chain var `transferHotspotStalePocBlocks`)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -45,6 +45,7 @@ const App = () => {
   }
 
   LogBox.ignoreLogs([
+    "Accessing the 'state' property of the 'route' object is not supported.",
     'Setting a timer',
     'Calling getNode() on the ref of an Animated component',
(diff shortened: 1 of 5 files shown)


matthewcarlreetz (approved):
✔️

PR #417, opened 2021-03-24
Create discovery mode with env vars for testing

matthewcarlreetz:
Discovery mode is currently faked by using a tab in place of a hotspot

You will need these env vars in order to test...
`DISCO_HOTSPOT_ADDRESS,DISCO_HOTSPOT_LAT,DISCO_HOTSPOT_LNG,DISCO_HOTSPOT_NAME`

Discovery mode will take the hotspot you select, clone it, and update these vars...
```
      const cloned = cloneDeep(propsHotspot)
      cloned.lat = lat
      cloned.lng = lng
      cloned.address = address
      cloned.name = name
```

If you have your own tab, you could use it for these variables, otherwise, I'll give you the ones I've been using.

I recommend testing on a real device as maps can be a pain to deal with on the simulator.
You can click on any nearby hotspot or "responding" hotspot do see some info about it.
You can currently run discovery on a hotspot 5 times a day, so use with caution.


src/features/hotspots/settings/discovery/DiscoveryMap.tsx
```diff
@@ -0,0 +1,326 @@
...
+  DiscoveryResponse,
+  DISCOVERY_DURATION_MINUTES,
+} from '../../../../store/discovery/discoveryTypes'
+import { hotspotsToFeatures } from '../../../../utils/mapUtils'
+import { useColors } from '../../../../theme/themeHooks'
+import { NetworkHotspot } from '../../../../store/networkHotspots/networkHotspotsSlice'
+import sleep from '../../../../utils/sleep'
+import usePrevious from '../../../../utils/usePrevious'
+
+const styleURL = 'mapbox://styles/petermain/ckjtsfkfj0nay19o3f9jhft6v'
```
tyler-whitman, line 34: Should we put this in a helper or const file so we can share it? Or maybe just export from the standard Map?

src/features/hotspots/settings/discovery/DiscoveryModeBegin.tsx
```diff
@@ -0,0 +1,108 @@
...
+        messageKey: 'discovery.begin.error.subtitle',
+        titleKey: 'discovery.begin.error.title',
+      })
+      onClose()
+    }
+    showError()
+  }, [showOKAlert, error, onClose, alertShown])
+
+  return (
+    <Box height={Math.min(640, hp(85))}>
```
tyler-whitman, line 60: Is this height to get right up to the top?

src/features/hotspots/settings/discovery/DiscoveryModeResults.tsx
```diff
@@ -0,0 +1,163 @@
...
+    isEqual,
+  )
+
+  const hotspotCoords = useMemo(() => {
+    return [hotspot.lng || 0, hotspot.lat || 0]
+  }, [hotspot.lat, hotspot.lng])
+
+  useEffect(() => {
+    const oneMileInDegrees = 1 / 69 // close enough => depends on your location. It's 68.7 at the equator and 69.4 at the poles, but yolo
+    const offset = 15 * oneMileInDegrees // TODO: 15 mile offset. Is this adequate?
```
tyler-whitman, line 62: So this is just to display network hotspots right? But some hotspots may still be found outside of the 15 mile range?

src/features/hotspots/settings/discovery/DiscoveryModeResults.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const showOverlay = async ({ name, lat, lng, address }: MapSelectDetail) => {
+    let distance = ''
+    if (hotspot.lat && hotspot.lng) {
+      const haversineDistance =
+        haversine({ lat: hotspot.lat, lng: hotspot.lng }, { lat, lng }) / 1000
+
+      const localizedDistance = usesMetricSystem
+        ? haversineDistance
+        : haversineDistance * 0.621371
+      const localizedUnit = usesMetricSystem ? 'km' : 'miles'
```
tyler-whitman, line 98: Do we need to translate these?

src/features/hotspots/settings/discovery/DiscoveryModeRoot.tsx
```diff
@@ -0,0 +1,197 @@
...
+        clearTimeout(requestInterval.current)
+      }
+      return
+    }
+
+    const reqId = requestId || selectedRequest?.id
+    if (!reqId) return
+    dispatch(fetchDiscoveryById({ requestId: reqId }))
+
+    requestInterval.current = setInterval(() => {
```
tyler-whitman, line 129: One thing to note here is this would stop if you put the app in the background.

We may want to use something like https://github.com/ocetnik/react-native-background-timer if we want it to keep running even if they minimize the app.

tyler-whitman (approved):
Looking great!

PR #418, opened 2021-03-25
Submit transactions through wallet-api

allenan:
This submits all transactions through wallet-api rather than to the api directly.

Note that we'll want to wait to merge this until the wallet api change has been deployed.

It also changes the notification list item to use the notification icon url that is supplied with each notification. I updated all of the icon url references in wallet api to use the new styles going forward.

### DIFF
--- a/src/hooks/useSubmitTxn.ts
+++ b/src/hooks/useSubmitTxn.ts
@@ -42,7 +42,7 @@ const useSubmitTxn = (): Submitter => {
     const serializedTxn = txn.toString()
 
     // submit to blockchain
-    const { data: apiResponse } = await postWallet('transactions', {
+    const apiResponse = await postWallet('transactions', {
       txn: serializedTxn,
     })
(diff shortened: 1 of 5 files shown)


src/hooks/useSubmitTxn.ts
```diff
@@ -18,7 +19,7 @@ import { useSelector } from 'react-redux'
 import activitySlice from '../store/activity/activitySlice'
 import { RootState } from '../store/rootReducer'
 import { useAppDispatch } from '../store/store'
-import { submitTransaction } from '../utils/appDataClient'
```
matthewcarlreetz, line 21: We should be able to remove `submitTransaction` from `appDataClient` now, right?

matthewcarlreetz (approved):
LGTM!

src/hooks/useSubmitTxn.ts
allenan: yes we can

tyler-whitman (approved):

Issue #424, opened 2021-03-25
show "never" instead of beginning of unix time


cokes518:
in diagnostic reports, if a value is 0, i.e. "last challenged", we show unix time instead of "never".

<img width="225" alt="image" src="https://user-images.githubusercontent.com/19331274/112551994-43a23900-8d7f-11eb-81e8-b5fab02afde0.png">

PR #426, opened 2021-03-27
Fix "Last Challenge" date shown as beginning of Unix time instead of "never"

MichaelBrew:
Fixes https://github.com/helium/hotspot-app/issues/424

If a hotspot does not have any challenges, export the "last challenge" as "never" in diagnostic reports instead of start of unix time (1970-01-01).

### DIFF
--- a/src/features/hotspots/settings/HotspotDiagnosticReport.tsx
+++ b/src/features/hotspots/settings/HotspotDiagnosticReport.tsx
@@ -213,7 +213,10 @@ const HotspotDiagnosticReport = ({ onFinished }: Props) => {
       natType: capitalize(diagnostics?.nat_type || ''),
       ip: capitalize(diagnostics?.ip || ''),
       height: info.height.toLocaleString(locale),
-      lastChallengeDate: format(fromUnixTime(info.lastChallengeTime), DF),
+      lastChallengeDate:
+        info.lastChallengeTime === 0
+          ? 'never'
+          : format(fromUnixTime(info.lastChallengeTime), DF),
       reportGenerated: format(fromUnixTime(info.currentTime), DF),
       gateway: address || '',
       hotspotMaker: onboardingRecord?.maker?.name || 'Unknown',


matthewcarlreetz (approved):
Looks good to me. Thanks!

src/features/hotspots/settings/HotspotDiagnosticReport.tsx
```diff
@@ -213,7 +213,10 @@ const HotspotDiagnosticReport = ({ onFinished }: Props) => {
       natType: capitalize(diagnostics?.nat_type || ''),
       ip: capitalize(diagnostics?.ip || ''),
       height: info.height.toLocaleString(locale),
-      lastChallengeDate: format(fromUnixTime(info.lastChallengeTime), DF),
+      lastChallengeDate:
+        info.lastChallengeTime === 0
+          ? 'never'
```
tyler-whitman, line 218: One thing we can do here is use translations for other languages. You only need to add the english string to the `src/locales/en.ts` file such as `never_challenged: 'never'` under the `hotspot_settings.diagnostics` object and then you can reference it using `t('hotspot_settings.diagnostics.never_challenged')`. We will then add the translations for other languages later.

tyler-whitman, line 218: One thing we can do here is use translations for other languages. You only need to add the english string to the `src/locales/en.ts` file such as `never_challenged: 'never'` under the `hotspot_settings.diagnostics` object and then you can reference it using `t('hotspot_settings.diagnostics.never_challenged')`. We will then add the translations for other languages later.

tyler-whitman (changes requested):
Thanks for fixing this bug! I mentioned a small improvement you can add.

src/features/hotspots/settings/HotspotDiagnosticReport.tsx
tyler-whitman: Wait nevermind, this gets sent to support so its not user facing. You are all good!

tyler-whitman (approved):
Thanks for fixing this!

PR #432, opened 2021-03-30
add support for rewards_v2

allenan:
adds support for `rewards_v2` transactions which the API should display the same as `rewards_v1`. 

depends on: https://github.com/helium/helium-js/pull/152

### DIFF
--- a/src/features/wallet/root/useActivityItem.tsx
+++ b/src/features/wallet/root/useActivityItem.tsx
@@ -8,6 +8,7 @@ import {
   PaymentV2,
   PendingTransaction,
   RewardsV1,
+  RewardsV2,
   TokenBurnV1,
   TransferHotspotV1,
 } from '@helium/http'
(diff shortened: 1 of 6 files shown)


src/features/wallet/root/useActivityItem.tsx
```diff
@@ -31,8 +31,12 @@ import { Colors } from '../../../theme/theme'
 import { getMakerName } from '../../../utils/stakingClient'
 import { RootState } from '../../../store/rootReducer'
 
+// TODO import from helium-js once released
+class RewardsV2 extends RewardsV1 {}
```
allenan, line 35: the helium-js PR basically does the same thing as this line since we're operating under the assumption that rewards v1 & v2 should contain the same data

tyler-whitman:
Changes look good, should we merge now or wait for rewards_v2 to land?

allenan:
> Changes look good, should we merge now or wait for rewards_v2 to land?

It would be nice to be able to test against a staging version of the API, I'll check in with @madninja for timing on that tomorrow

tyler-whitman (approved):

PR #438, opened 2021-03-31
Add follow toggle to hotspot details. Show followed hotspots in list.

### LINKED ISSUE #422: Favorite Hotspot: Browsing Map
Another entry point to favorite a Hotspot is when a user is browsing the network map.
Tap on a dot, see the Hotspot details pop up, press the heart.

matthewcarlreetz:
Closes #422 
Dependent on Wallet API PR https://github.com/helium/wallet-api/pull/25

https://user-images.githubusercontent.com/1760821/113166067-6cf41680-9210-11eb-8b2d-be44b188baaf.mov


tyler-whitman (changes requested):
Looking good! Ill test it when we merge the api changes

src/features/hotspots/settings/HotspotsDiagnosticOptions.tsx
```diff
@@ -16,14 +17,19 @@ import { useColors } from '../../../theme/themeHooks'
...
-  const {
-    connectedHotspot: { firmware },
-  } = useSelector((state: RootState) => state)
+  const firmware = useSelector(
+    (state: RootState) => state.connectedHotspot.firmware,
+  )
+  const owner = useSelector(
+    (state: RootState) => state.connectedHotspot.details?.owner,
+  )
+  const { result: address } = useAsync(getSecureItem, ['address'])
```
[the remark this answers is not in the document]
matthewcarlreetz: I suppose it could go either way, but I'd rather pull it directly from `secureAccount`

src/store/hotspots/hotspotsSlice.ts
```diff
@@ -34,46 +67,87 @@ const hotspotSorters: Record<HotspotSort, HotspotSorter> = {
...
+
+    const ownedAddresses = hotspots.map((h) => h.address)
+    const followingAddresses = followedHotspots.map((h) => h.address)
+    const unOwnedAddresses = followingAddresses.filter(
+      (fa) => !ownedAddresses.includes(fa),
+    )
+
+    const allAddresses = [...unOwnedAddresses, ...ownedAddresses]
+
+    const unOwnedHotspots = await Promise.all(
```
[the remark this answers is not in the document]
matthewcarlreetz: I was torn on it. The sync status was always slightly off which was annoying to me. I'm sure it's not a big deal, but I'd rather have it be right. Thoughts?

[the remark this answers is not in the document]
tyler-whitman: Yeah that sounds good

src/features/hotspots/settings/HotspotsDiagnosticOptions.tsx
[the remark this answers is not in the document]
tyler-whitman: Its nice to be able to quickly mock accounts when its all funneled through one function is my only reasoning

src/App.tsx
```diff
@@ -158,6 +158,14 @@ const App = () => {
     }
   }, [fetchDataStatus, isBackedUp, isRestored])
 
+  useEffect(() => {
```
matthewcarlreetz, line 161: This isn't related to this PR, but it's been driving me crazy.

matthewcarlreetz: It seems to work fine, not sure if 5s is the right number though.

allenan:
noticed a followed hotspot shows 0 HNT on the list screen

image

src/App.tsx
tyler-whitman: Hmm yeah I think mine is usually like a second or two. I guess if we want to do fixed maybe 2 seconds?

tyler-whitman: Oh or this is only to handle the case it takes longer than 5 seconds?

allenan:
This may be handled in the other PR in which case disregard, but if an account has followed a hotspot but has no hotspots of its own, it should be able to see them on the list screen:

image

tyler-whitman:
Should we add an empty view when you have no followed hotspot and you use the filter? Maybe some text or something?

tyler-whitman:
> noticed a followed hotspot shows 0 HNT on the list screen

This seems to happen on right after you follow it. If you re launch the app it fixes.

tyler-whitman (approved):
Working well for me besides what @allenan pointed out.

tyler-whitman:
I also found one more thing, seems like if you go to the map and go back to your followed list and then quickly tap the followed hotspot in the list maybe it doesnt have enough time to load the coords and it goes to a random location I think because of invalid cords. I see this errror

```
Warning: Failed prop type: Invalid prop `bounds.ne[0]` of type `string` supplied to `Camera`, expected `number`.
    in Camera (at Map.tsx:336)
    in Map
    in Map (at HotspotsView.tsx:245)
```

src/App.tsx
matthewcarlreetz: Yeah, this is so that at worst it will show 5 seconds.

matthewcarlreetz:
> I also found one more thing, seems like if you go to the map and go back to your followed list and then quickly tap the followed hotspot in the list maybe it doesnt have enough time to load the coords and it goes to a random location I think because of invalid cords. I see this errror
> 
> ```
> Warning: Failed prop type: Invalid prop `bounds.ne[0]` of type `string` supplied to `Camera`, expected `number`.
>     in Camera (at Map.tsx:336)
>     in Map
>     in Map (at HotspotsView.tsx:245)
> ```


Oh, good catch! That's because the lat/lng coming from the wallet API are strings. They get replaced eventually by their  block versions.

PR #447, opened 2021-04-02
Feature/follow onboarding and discovery

### LINKED ISSUE #420: Favorite Hotspot: Discovery Mode
Allow users to favorite a Hotspot after a Discovery Mode session.

Users may want to do this to follow Hotspot earnings of nearby Hotspots.

Obviously blocked by Discovery Mode having launched.

### LINKED ISSUE #421: Favorite Hotspot: Onboarding Flow
For users that are hosting Hotspots, they may want to favorite/follow straight from the onboarding flow.

On the screen that says they don't own this hotspot, give users the ability to favorite the Hotspot they just tried to onboard.

matthewcarlreetz:
Closes #420 
Closes #421

https://user-images.githubusercontent.com/1760821/113435752-72d52d80-93b1-11eb-81e7-8927ee098627.mov

https://user-images.githubusercontent.com/1760821/113436047-fe4ebe80-93b1-11eb-81ef-7e13c9722938.mov

### DIFF
--- a/src/components/FollowButton.tsx
+++ b/src/components/FollowButton.tsx
@@ -12,9 +12,15 @@ import { Theme } from '../theme/theme'
 
 type Props = BoxProps<Theme> & {
   address: string
+  colors?: { following: string; notFollowing: string }
   handleChange?: (following: boolean) => void
 }
-const HotspotDetailsHandle = ({ address, handleChange, ...props }: Props) => {
+const HotspotDetailsHandle = ({
+  address,
+  colors,
+  handleChange,
+  ...props
+}: Props) => {
   const { grayPurple, followPurple } = useColors()
   const dispatch = useAppDispatch()
   const [following, setFollowing] = useState(false)
@@ -26,11 +32,11 @@ const HotspotDetailsHandle = ({ address, handleChange, ...props }: Props) => {
     setFollowing(!!followedHotspots[address])
   }, [followedHotspots, address])
 
-  const color = useMemo(() => (following ? followPurple : grayPurple), [
-    following,
-    followPurple,
-    grayPurple,
-  ])
+  const color = useMemo(() => {
+    if (!colors) return following ? followPurple : grayPurple
+
+    return following ? colors.following : colors.notFollowing
+  }, [colors, following, followPurple, grayPurple])
 
   const toggleFollowing = useCallback(() => {
     setFollowing(!following)
--- a/src/components/FollowButton.tsx
+++ b/src/components/FollowButton.tsx
@@ -1,0 +1,56 @@
+/* eslint-disable react/jsx-props-no-spreading */
+import React, { memo, useCallback, useEffect, useMemo, useState } from 'react'
+import { useSelector } from 'react-redux'
+import Follow from '@assets/images/follow.svg'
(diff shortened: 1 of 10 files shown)


src/locales/en.ts
```diff
@@ -898,6 +901,7 @@ export default {
       result_time: 'Time of Results',
       searching: 'Searching',
       distance: '{{distance}} {{unit}} away',
+      added_to_favorites: 'Added to Favorites',
```
allenan, line 904: should we call this "follow" instead of "favorite"? something like "Added to followed hotspots" or "Followed hotspot"?

src/components/FollowButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+const HotspotDetailsHandle = ({
+  address,
+  colors,
+  handleChange,
+  ...props
+}: Props) => {
+  const { grayPurple, followPurple } = useColors()
+  const dispatch = useAppDispatch()
+  const [following, setFollowing] = useState(false)
```
allenan, line 26: is it necessary to have both local state and redux state for following?

src/features/hotspots/setup/NotHotspotOwnerError.tsx
```diff
@@ -1,44 +1,182 @@
...
+          <Text
+            marginLeft="l"
+            style={hotspotNameStyle}
+            variant="medium"
+            fontSize={16}
+            flex={1}
+          >
+            {startCase(name)}
+          </Text>
+          <FollowButton
```
allenan, line 148: just wondering if it's obvious enough that only the flag here is pressable? it sort of looks like the whole box is a button. Though I know we're trying to keep all the follow logic contained within the `FollowButton` component which is great btw. I guess with the animation maybe that points out the flag enough. we can see what happens with users

allenan (approved):
works well and I like how compartmentalized the follow logic is. down to get this onto TF 👍

src/components/FollowButton.tsx
matthewcarlreetz: The round trip to the server takes a bit, that's why I have local state alongside redux state.

allenan:
it looks like the "nearest" and "followed" sort options use the same icon, maybe we could use a globe or something for "nearest"? we can address in a separate PR since there probably isn't a design for it

image

src/locales/en.ts
matthewcarlreetz: oh, yeah, good catch.

matthewcarlreetz:
> it looks like the "nearest" and "followed" sort options use the same icon, maybe we could use a globe or something for "nearest"? we can address in a separate PR since there probably isn't a design for it
> 
> image

Oh, I didn't even realize that. We have varying terminology on the designs in Figma. I'll chat with Pete about it but probably will need to wait for a future PR.

src/features/hotspots/setup/NotHotspotOwnerError.tsx
matthewcarlreetz: yeah, we could always add children to the FollowButton, but let's wait and get feedback.

tyler-whitman (approved):

PR #452, opened 2021-04-06
Assert V2

tyler-whitman:
Reworks the assert flow to no longer require hotspot pairing. This introduces the Assert V2 transaction to the app.

This is still pending the API change for https://github.com/helium/blockchain-http/issues/208 and a new release of Helium JS.


src/utils/polyfill.js
```diff
@@ -0,0 +1,8 @@
+import { Platform } from 'react-native'
+
+// noinspection JSConstantReassignment
```
matthewcarlreetz, line 3: what's this here for?

src/components/HotspotConfigurationPicker.tsx
```diff
@@ -0,0 +1,228 @@
...
+    onGainUpdated(antenna.gain)
+    setGain(
+      antenna.gain.toLocaleString(locale, {
+        maximumFractionDigits: 1,
+        minimumFractionDigits: 1,
+      }),
+    )
+  }
+
+  const showElevationInfo = () =>
```
matthewcarlreetz, line 84: Is this placeholder?

src/components/HotspotConfigurationPicker.tsx
```diff
@@ -0,0 +1,228 @@
...
+      borderRadius="m"
+      marginVertical="l"
+      borderWidth={outline ? 1 : 0}
+      borderColor="grayLight"
+    >
+      <HeliumActionSheet
+        title={t('antennas.onboarding.select')}
+        prefixVariant="regular"
+        prefixFontSize={14}
+        data={Object.values(Antennas).map((a) => ({
```
matthewcarlreetz, line 161: Wouldn't hurt to move this data out to a `useMemo`

src/features/hotspots/settings/ReassertAddressSearch.tsx
```diff
@@ -99,7 +101,7 @@ const ReassertAddressSearch = ({ onSelectPlace }: Props) => {
   )
 
   return (
-    <Box height={hp(75)} padding="none" paddingTop="lx">
+    <Box height={600} padding="none" paddingTop="lx">
```
matthewcarlreetz: Is this too tall for some phones?

matthewcarlreetz (commented):
Code is looking good. Excited to try it out tonight.

matthewcarlreetz (approved):
Worked great!

src/utils/polyfill.js
tyler-whitman: This is for H3, it needs a shim to work but it doesnt actually use the document so it works great. I added that noinspect to ignore some lint error.

src/components/HotspotConfigurationPicker.tsx
tyler-whitman: Its the alerts that pop up when you hit the i icons near the elevation and gain text. Shows an alert with some info.

src/components/HotspotConfigurationPicker.tsx
tyler-whitman: Done

src/features/hotspots/settings/ReassertAddressSearch.tsx
tyler-whitman: I think it should be ok, when we use the hp I notice things sometimes differ but fixed pixel is supposed to be density independent so it should always be the same height on any device. Let me try it on a small phone though

tyler-whitman: It is the same height but its taller than the device on the SE lol. Ill make it smaller for small devices

PR #453, opened 2021-04-06
hotspot checklist feature flag

tyler-whitman:


src/features/hotspots/checklist/HotspotChecklist.tsx
```diff
@@ -27,18 +27,23 @@ const HotspotChecklist = ({ hotspot, witnesses }: Props) => {
...
+    (state: RootState) => state.features.checklistEnabled,
+  )
 
   useEffect(() => {
-    dispatch(fetchChecklistActivity(hotspot.address))
+    if (checklistEnabled) {
+      dispatch(fetchChecklistActivity(hotspot.address))
+    }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
```
allenan, line 48: does `checklistEnabled` need to be included in the hook's dependencies?

src/store/features/featuresSlice.ts
```diff
@@ -3,15 +3,18 @@ import { getWallet } from '../../utils/walletClient'
 
 export type FeaturesState = {
   discoveryEnabled: boolean
+  checklistEnabled: boolean
```
allenan, line 6: did you test to make sure that when the feature flag is true, checklist gets enabled?

src/features/hotspots/checklist/HotspotChecklist.tsx
tyler-whitman: Yeah good point, since before that was only called at start but if the flag gets enabled we should call it. Ill add all the dependencies now

src/store/features/featuresSlice.ts
tyler-whitman: Yeah I mocked it locally and it worked

allenan (approved):

PR #467, opened 2021-04-12
only load checklist when expanded and fix cropped text

### LINKED ISSUE #450: Hotspot Checklist clipping
When viewing the app in Chinese, the characters are clipped at the top. Need to add some padding there.

image

### LINKED ISSUE #459: Show/Hide Checklist toggle
Add back Hotspot Checklist to app - but only show on toggle.
Default should be "off".

We'll also need to turn the feature flag back on.

tyler-whitman:
closes #450 
closes #459 

The checklist is still behind a feature flag that will need to be turned on to show up. You can mock it in the FeaturesSlice if you want to test.


matthewcarlreetz (approved):
LGTM!

PR #471, opened 2021-04-15
Feature/419 text search for hotspot

### LINKED ISSUE #419: Favorite Hotspot Entry: Search
Add search functionality to Hotspot app to search the network for a Hotspot name.
Also add ability to filter for a Hotspot within a user's wallet (we'll need to identify an owned Hotspot in the search results)

@PeterMain if you can add your design input in figma: WIP: Favorite Hotspot -> Entry Point 1.

matthewcarlreetz:
Closes #419 

Made lots of changes, so please be sure to test thoroughly.

https://user-images.githubusercontent.com/1760821/114889609-e0308780-9dcf-11eb-9460-0b53825300b5.mov


src/components/SearchInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+  initialValue?: string
+}
+const HotspotSearch = ({ onSearch, initialValue = '', ...boxProps }: Props) => {
+  const { t } = useTranslation()
+  const [searchTerm, setSearchTerm] = useState(initialValue)
+
+  useEffect(() => setSearchTerm(initialValue), [initialValue])
+
+  const debouncedSearchTermHandler = useDebouncedCallback((term: string) => {
+    onSearch(term)
```
tyler-whitman, line 27: Should hotspot names with spaces also work? It seems to only work if I type dashes?

tyler-whitman: Another thing, when I am on the "My Hotspots" tab, the hint text made me feel like I could also search a location to filter my own hotspots. Like if I type city names I would see my hotspots in that city.

src/components/SearchInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+        variant="light"
+        placeholder={t('hotspots.search.placeholder')}
+        onChangeText={setSearchTerm}
+        value={searchTerm}
+        keyboardAppearance="dark"
+        autoCorrect={false}
+        autoCompleteType="off"
+        autoCapitalize="none"
+        blurOnSubmit={false}
+        returnKeyType="search"
```
tyler-whitman, line 64: I guess this return key type doesnt dismiss the keyboard? Can we maybe use `done` or when they hit it maybe we dismiss the keyboard?

We could also add a little X icon in the search input to stop search?

src/features/hotspots/root/HotspotsView.tsx
```diff
@@ -51,18 +57,23 @@ type Props = {
...
-  location,
+  location: propsLocation,
 }: Props) => {
   const navigation = useNavigation()
   const { t } = useTranslation()
   const dispatch = useDispatch()
   const colors = useColors()
-
-  const [listIsDismissed, setListIsDismissed] = useState(!!startOnMap)
+  const [viewState, setViewState] = useState<ViewState>(
```
tyler-whitman, line 72: Nice this cleans it up a bit, this view keeps getting more complicated haha

src/features/hotspots/root/HotspotsView.tsx
```diff
@@ -105,66 +154,60 @@ const HotspotsView = ({
...
         setSelectedHotspot(ownedHotspots[0])
       } else {
         focusClosestHotspot()
       }
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [listIsDismissed, networkHotspots])
-
-  useEffect(() => {
-    if (startOnMap) {
```
tyler-whitman, line 119: Some of this was for when you navigate here from the "view hotspots around me" so that it selects the closest hotspot to you and shows it first. Did you change how that works?

src/features/hotspots/root/HotspotsView.tsx
```diff
@@ -105,66 +154,60 @@ const HotspotsView = ({
...
+    [updateViewState],
+  )
 
-  const handleDismissList = useCallback(() => {
-    detailsRef.current?.present()
-    setListIsDismissed(true)
-  }, [])
+  const handleListDismiss = useCallback(() => {
+    if (viewState === 'details') return
+    updateViewState('map')
```
tyler-whitman, line 181: I notice now that when I just swipe down the list card, I end up on the map with my last selected hotspot but the details card never shows.

src/utils/appDataClient.ts
```diff
@@ -44,6 +44,14 @@ export const getHotspots = async () => {
   return newHotspotList.takeJSON(1000)
 }
 
+export const searchHotspots = async (searchTerm: string) => {
+  const address = await getAddress()
+  if (!address) return []
+
+  const newHotspotList = await client.hotspots.search(searchTerm)
+  return newHotspotList.takeJSON(1000)
```
tyler-whitman, line 52: does this API use a cursor?

tyler-whitman (changes requested):

src/features/hotspots/root/HotspotsView.tsx
tyler-whitman: Hmm it seems like that is no longer working on Android, but it seems related to previous location changes

src/utils/appDataClient.ts
matthewcarlreetz: Actually, no, there isn't one.

src/components/SearchInput.tsx
matthewcarlreetz: oh, ok, I should be able to add that.

src/utils/appDataClient.ts
tyler-whitman: Ok cool, looks good then

src/features/hotspots/root/HotspotsView.tsx
matthewcarlreetz: My intention wasn't to change how it works, just to bring all of that logic into one place. As far as I can tell everything is working as it was.

matthewcarlreetz: I updated it so that when you dismiss the list it will bring the user to their location if they give permission otherwise it will bring them to one of their hotspot's locations.

src/features/hotspots/root/HotspotsView.tsx
matthewcarlreetz: We decided to change it up so that it doesn't auto-select a hotspot for you.

tyler-whitman (approved):

PR #473, opened 2021-04-16
Update hotspots list styling

matthewcarlreetz:
This mostly matches figma, but there are some design pieces that still need to be finalized. Decided to merge this progress and will make another pass at it later.


src/features/hotspots/root/HotspotsTicker.tsx
```diff
@@ -0,0 +1,67 @@
...
+  )
+  const blockTime = useSelector(
+    (state: RootState) => state.heliumData.blockTime,
+  )
+
+  useEffect(() => {
+    dispatch(fetchCurrentOraclePrice())
+    dispatch(fetchPredictedOraclePrice())
+    dispatch(fetchStats())
+  }, [dispatch])
```
tyler-whitman, line 38: Do we want to put this on a 5 min update interval? That is what the rest of the page is using for rewards info. If you check in `WelcomeOverview` there is a date that gets updated every 5 min, you could just pass it down and add it to the dependencies so this refreshes too?

tyler-whitman: Or we could maybe make it like 15 or 30 minutes and do it within this component?

tyler-whitman (approved):

src/features/hotspots/root/HotspotsTicker.tsx
matthewcarlreetz: Maybe I should do it every time it focuses?

tyler-whitman: Hmm just dont want to hammer the API. I think every 5 min is enough

matthewcarlreetz: What if I combine them both? On focus, if it's been more than 5 minutes.

tyler-whitman: That sounds good

PR #477, opened 2021-04-17
Hotspot Tab Fixes

matthewcarlreetz:
Show settings cog when clicking on hotspot from map.
Nav to proper location when choosing location from search.
Change ticker easing to linear
Couple eslint/husky fixes


tyler-whitman (approved):
Looks good

Issue #134, opened 2021-02-01
Allow QR code scanning of payment_v2 transactions


rawrmaan:
When using the Send feature, allow the user to scan `payment_v2` transactions as a JSON object with the keys being payee addresses and the values being amounts in bones.

This allows large deployers to easily make bulk payments from their master accounts, avoiding additional custodial risk of either having to use the CLI or trust a 3rd party to forward a `payment_v2` on their behalf.

This will require a massive QR code, but luckily QR codes can sure up to `2953` bytes: https://stackoverflow.com/questions/11065415/how-much-data-information-can-we-save-store-in-a-qr-code

PR #479, opened 2021-04-18
[#134] Allow QR code scanning of payment_v2 transactions

MichaelBrew:
Feature request: https://github.com/helium/hotspot-app/issues/134

### Summary
This PR updates the "Send" flow to accommodate QR-codes with multiple transfers. In this context, a "transfer" can represent a payment, DC burn, or hotspot token transfer. It now supports QR codes with the following stringified JSON structure:
```json
{
  "type": "payment",
  "payees": [
    {
      "address": "...",
      "amount": 10
    }
  ]
}
```
Those transfers are rendered sequentially within the `SendView` screen. There is currently no option to add or remove individual transfers from the `SendView` screen after initializing from a QR scan.

### Component Hierarchy
Transfers are passed from `ScanView` to `SendView`, which then renders `SendForm`, which then renders individual `SendDetailsForm` components for each transfer. This new `SendDetailsForm` component is responsible for capturing address/amount/memo input and running associated validators (e.g. ETH address mapping, HNT balance computation, fee calculation, etc) and propagating those changes up to the parent `SendView` component.

Snip20210424_47

### Testing
TBD


src/features/wallet/send/SendTransferForm.tsx
```diff
@@ -0,0 +1,420 @@
...
+  isSeller?: boolean
+  lastReportedActivity?: string
+  onScanPress: () => void
+  sendTransfer: SendTransfer
+  transferData?: Transfer
+  type: SendType
+  updateTransfer: (transferId: number, updates: any) => void
+}
+
+const SendTransferForm = ({
```
tyler-whitman, line 41: Thank you for explaining your refactor any why it was necessary. I agree with your reasoning and structure of the components, nice idea to get multiple payment address and amounts supported. 

One thing that this naming is a little confusing with is that this form also handles Hotspot Transfer transaction creation and submission (transferring hotspots between accounts). With your refactor you have made this new object which you call SendTransfers and I think the naming conflicts and makes it a little confusing.

My suggestion is maybe we come up with a new name for these payment / payee pairs. Maybe something like `SendDetailsForm`? Open to other suggestions as well but the Transfer word is the one that makes it confusing.

Then we would call the objects `SendDetails`? Maybe @allenan or @matthewcarlreetz have some input on this naming too.

src/features/wallet/send/SendTransferForm.tsx
```diff
@@ -0,0 +1,420 @@
...
+  const [amount, setAmount] = useState<string>(sendTransfer.amount)
+  const [balanceAmount, setBalanceAmount] = useState<Balance<NetworkTokens>>(
+    sendTransfer.balanceAmount,
+  )
+  const [dcAmount, setDcAmount] = useState<string>(sendTransfer.dcAmount)
+  const [memo, setMemo] = useState<string>(sendTransfer.memo)
+  const [fee, setFee] = useState<Balance<NetworkTokens>>(sendTransfer.fee)
+
+  useEffect(() => {
+    updateTransfer(sendTransfer.id, {
```
tyler-whitman, line 75: Looks like `SendTransfer` object has a string id, but the `updateTransfer` functions id parameter is a number.

src/features/wallet/send/SendTransferForm.tsx
```diff
@@ -0,0 +1,420 @@
...
+      const maxAmountStr = maxAmount.toString(8, {
+        decimalSeparator,
+        groupSeparator,
+        showTicker: false,
+      })
+      setAmount(maxAmountStr)
+    }
+  }
+
+  const renderLockedPaymentForm = () => [
```
tyler-whitman, line 185: This shouldn't return an array, it should use `(` and `)` instead of `[` and `]`

src/features/wallet/send/SendTransferForm.tsx
```diff
@@ -0,0 +1,420 @@
...
+    <LockedField label={t('send.address.label')} value={address} />,
+    <LockedField
+      label={t('send.amount.label')}
+      value={amount}
+      footer={amount ? <FeeFooter fee={fee} /> : undefined}
+      bottom
+    />,
+  ]
+
+  const renderLockedBurnForm = () => [
```
tyler-whitman, line 195: This shouldn't return an array, it should use `(` and `)` instead of `[` and `]`

src/features/wallet/send/SendTransferForm.tsx
```diff
@@ -0,0 +1,420 @@
...
+    <LockedField
+      label={t('send.amount.label')}
+      value={amount}
+      footer={amount ? <FeeFooter fee={fee} /> : undefined}
+    />,
+    <LockedField label={t('send.dcAmount.label')} value={dcAmount} />,
+    <LockedField label={t('send.memo.label')} value={memo} bottom />,
+  ]
+
+  const renderPaymentForm = () => [
```
tyler-whitman, line 206: This shouldn't return an array, it should use `(` and `)` instead of `[` and `]`

src/features/wallet/send/SendTransferForm.tsx
```diff
@@ -0,0 +1,420 @@
...
+          <Text fontSize={12} color="primaryMain">
+            {t('send.sendMax')}
+          </Text>
+        </TouchableOpacityBox>
+      }
+      footer={amount ? <FeeFooter fee={fee} /> : undefined}
+    />,
+  ]
+
+  const renderBurnForm = () => [
```
tyler-whitman, line 239: This shouldn't return an array, it should use `(` and `)` instead of `[` and `]`

src/features/wallet/send/SendTransferForm.tsx
```diff
@@ -0,0 +1,420 @@
...
+    />,
+    <InputField
+      defaultValue={memo}
+      onChange={setMemo}
+      label={t('send.memo.label')}
+      placeholder={t('send.memo.placeholder')}
+    />,
+  ]
+
+  const renderSellerTransferForm = () => [
```
tyler-whitman, line 286: This shouldn't return an array, it should use `(` and `)` instead of `[` and `]`

src/features/wallet/send/SendTransferForm.tsx
```diff
@@ -0,0 +1,420 @@
...
+      defaultValue={amount}
+      onChange={setFormAmount}
+      value={amount}
+      numberOfLines={2}
+      label={t('send.amount.label_transfer')}
+      placeholder={t('send.amount.placeholder_transfer')}
+    />,
+  ]
+
+  const renderBuyerTransferForm = () => [
```
tyler-whitman, line 320: This shouldn't return an array, it should use `(` and `)` instead of `[` and `]`

src/features/wallet/send/SendTransferForm.tsx
```diff
@@ -0,0 +1,420 @@
...
+type Props = {
+  account?: Account
+  isLocked: boolean
+  isSeller?: boolean
+  lastReportedActivity?: string
+  onScanPress: () => void
+  sendTransfer: SendTransfer
+  transferData?: Transfer
+  type: SendType
+  updateTransfer: (transferId: number, updates: any) => void
```
tyler-whitman, line 38: can we type the updates here?

src/features/wallet/send/SendView.tsx
```diff
@@ -420,33 +372,43 @@ const SendView = ({ scanResult, sendType, hotspot, isSeller }: Props) => {
     }
   }
 
-  const constructTxn = async () => {
+  const constructTxns = async () => {
     if (type === 'payment') {
-      return makePaymentTxn(balanceAmount.integerBalance, address, getNonce())
+      return Promise.all(
```
tyler-whitman, line 377: So your goal of allowing multiple address / payee fields in a QR for payment was to be able to create a QR that can send multiple addresses different amounts correct?

The logic here is not the correct way to go about that. For a payment with multiple payees and amounts, we need to pass them all into one payment_v2 transaction. This is something we currently support on the blockchain but haven't done in the app yet. 

Instead of creating multiple transaction like you have done, we need to create one transaction and fill the payee and amounts correctly in the payments array of that single payment_v2 transaction.

You will need to revert this change and make updates to https://github.com/helium/hotspot-app/blob/main/src/utils/transactions.ts#L26 to allow multiple payee's and amounts.

src/features/wallet/send/SendView.tsx
```diff
@@ -420,33 +372,43 @@ const SendView = ({ scanResult, sendType, hotspot, isSeller }: Props) => {
...
+      )
     }
 
     if (type === 'dc_burn') {
-      return makeBurnTxn(
-        balanceAmount.integerBalance,
-        address,
-        getNonce(),
-        memo,
+      return Promise.all(
```
tyler-whitman, line 389: Is supporting multiple burns in the same QR something you wanted to do? If so maybe we can talk about that. Submitting rapid transactions like this will probably fail and we may need to make some changes to support burns in this way. See my comment about about payments.

My suggestion would be that we only allow multiple payee / amounts for payment transactions at this moment.

src/features/wallet/send/SendView.tsx
```diff
@@ -420,33 +372,43 @@ const SendView = ({ scanResult, sendType, hotspot, isSeller }: Props) => {
...
   }
 
   const handleSubmit = async () => {
     try {
-      const txn = await constructTxn()
-      if (txn) {
-        await submitTxn(txn)
-      }
+      const txns = await constructTxns()
+      await Promise.all(txns.map(async (txn: any) => submitTxn(txn)))
```
tyler-whitman, line 411: After addressing my other comments about this, we will need to change this back to just submitting one transaction.

src/features/wallet/send/SendView.tsx
```diff
@@ -151,53 +148,43 @@ const SendView = ({ scanResult, sendType, hotspot, isSeller }: Props) => {
...
-  }, [type, balanceAmount])
-
   // validate transaction
   useEffect(() => {
-    const isValidAddress = Address.isValid(address)
-
     if (type === 'transfer') {
+      const { fee } = sendTransfers[0]
       if (isSeller) {
-        setIsValid(isValidAddress && (hasValidActivity || false))
```
tyler-whitman, line 200: This breaks the send button for the Hotspot Transfer form. Why did you remove it? To get to this flow you need to go to a hotspot, click settings, and then initiate a transfer of that hotspot.

src/features/wallet/send/SendView.tsx
```diff
@@ -217,64 +204,55 @@ const SendView = ({ scanResult, sendType, hotspot, isSeller }: Props) => {
...
       }
     } else {
-      const totalTxnAmount = balanceAmount.plus(fee)
-      // TODO balance compare/greater than/less than
-      const hasBalance =
-        totalTxnAmount.integerBalance <= (account?.balance?.integerBalance || 0)
+      let hasBalance = true
+      let isValidAddress = true
+      let isValidBalanceAmount = true
+      sendTransfers.forEach(({ address, balanceAmount, fee }) => {
```
tyler-whitman, line 210: This is validating the amount for each payment in the array, but it should be adding up all the amounts and validating that the user has enough balance to pay the total

tyler-whitman (changes requested):
Thanks for submitting these changes! I think you have a good start and this is a pretty big refactor. My main concern is that it is breaking the existing Hotspot Transfer flow, we will need to test Burns, Transfers, Payments, and QR pretty heavily after we get this into a good state.

Let me know if you have any questions about my feedback or want to jump on a call and discuss.

src/features/wallet/send/SendTransferForm.tsx
MichaelBrew: Great suggestion, thank you! I didn't fully realize what the "transfer" was in the context of this Send flow, so now that you explain it, it makes total sense. Just renamed instances of `SendTransfer` to `SendDetails`.

src/features/wallet/send/SendTransferForm.tsx
MichaelBrew: For this comment (and the related ones within this file), is this stylistic feedback? Or is there a functional purpose? Given all form elements are nested within a `<Box />` in the component's return value, I figured we didn't need to wrap them again within another `<Box />` (or React fragment) in these internal helpers.

src/features/wallet/send/SendView.tsx
MichaelBrew: Don't think I purposefully removed this, was more so a casualty of multiple iterations 🙂  added back https://github.com/helium/hotspot-app/pull/479/commits/a97ebb18637f53dda8d0c224d3c20c6293866a70

src/features/wallet/send/SendView.tsx
MichaelBrew: Great catch, fixed here https://github.com/helium/hotspot-app/pull/479/commits/c37e5c463b597e613e27eaaa88ec71dacdd197e6

src/features/wallet/send/SendView.tsx
```diff
@@ -334,7 +281,7 @@ const SendView = ({ scanResult, sendType, hotspot, isSeller }: Props) => {
       Alert.alert(t('generic.error'), t('send.error'))
       throw new Error('failed to create seller TransferHotspotV1 transaction')
     }
-    const transfer = createTransfer(
+    const transfer = await createTransfer(
```
MichaelBrew, line 284: Noticed this wasn't previously awaiting the promise from `createTransfer`. Not exactly sure what the user-facing impact is here because this function returns undefined anyway (gets past the `!transfer` check because `transfer` is actually a promise).

MichaelBrew:
@tyler-whitman thanks for the thoughtful review - have pushed several changes based on feedback, including the limitation of multiple "send" actions to only the "payment" send type, and submitting those payments in a single transaction (https://github.com/helium/hotspot-app/pull/479/commits/ef61ccb775622351ce1eaf7d90f3d98e13e70d73).

As indicated by the "TBD" label in the PR description, would love some guidance on how to test these changes. This is my first (non-one-liner) PR to the repo so still learning best practices 🙂  Is there any sort of automated test suite that can be run against this branch? Or do we need to manually test? Is there a sandbox/dev/staging environment? Or do things have to be tested against production?

src/features/wallet/send/SendTransferForm.tsx
matthewcarlreetz: I actually really like this syntax, my concern is that I'm pretty sure it will create [this](https://reactjs.org/docs/lists-and-keys.html#keys) warning because arrays need to have keys.

Typically we've been just wrapping in fragments like this 
```
const renderLockedPaymentForm = () => (
  <>
    <Component />
    <OtherComponent />
  </>
)

but, you could just give each component a key and keep this current style.

Comment (tyler-whitman) [reply src/features/wallet/send/SendTransferForm.tsx]: I think for fixed outputs a fragment is the expected element, arrays should be used for more dynamic content. More info on fragments can be found here: https://reactjs.org/docs/fragments.html

File: src/features/wallet/send/SendView.tsx (hunk 15 above)
Comment (tyler-whitman) [reply src/features/wallet/send/SendView.tsx:284]: Oh good catch!

File: src/features/wallet/send/SendView.tsx (hunk 10 above)
Comment (tyler-whitman) [reply src/features/wallet/send/SendView.tsx]: The changes you made look good!

File: src/features/wallet/send/SendView.tsx (hunk 14 above)
Comment (tyler-whitman) [reply src/features/wallet/send/SendView.tsx]: This is still a little incorrect, on line 230 you have `totalSendAmount = totalSendAmount.plus(balanceAmount.plus(fee))` this should just be `totalSendAmount = totalSendAmount.plus(balanceAmount)` and then at the end outside the loop you just add the fee once, `totalSendAmount = totalSendAmount.plus(fee)`.

The fee is the transaction fee on the blockchain and its only one fee per transaction, which we are now using just one paymentV2 transcation with multiple payees and amounts.

Comment (tyler-whitman) [reply src/features/wallet/send/SendView.tsx]: So in the end you add up, all the payment amount plus the one transaction fee

**Comment by tyler-whitman** [comment]
> @tyler-whitman thanks for the thoughtful review - have pushed several changes based on feedback, including the limitation of multiple "send" actions to only the "payment" send type, and submitting those payments in a single transaction ([ef61ccb](https://github.com/helium/hotspot-app/commit/ef61ccb775622351ce1eaf7d90f3d98e13e70d73)).
> 
> As indicated by the "TBD" label in the PR description, would love some guidance on how to test these changes. This is my first (non-one-liner) PR to the repo so still learning best practices 🙂 Is there any sort of automated test suite that can be run against this branch? Or do we need to manually test? Is there a sandbox/dev/staging environment? Or do things have to be tested against production?

Thanks for all the fixes, this is looking really good now.

Testing is very manual for us as of now. We are just two main engineers working on the mobile app and haven't had time to establish an automated test suite. We also currently test against production. There is a testnet for validator work but we haven't added support for it in the app yet. 

I would recommend generating some QRs with the new data format you added and verify the details fill correctly. I will do some testing from my side on the actual transactions. You can send HNT between accounts and lose only a small amount to fees (usually $0.35 or so) which is what I do most of the time. I will make sure to test some payments and transfers once we have it in a ready to test state (which is close, I left two comments open).

Thank you again for submitting these changes!

File: src/features/wallet/send/SendTransferForm.tsx (hunk 3 above)
Comment (MichaelBrew) [reply src/features/wallet/send/SendTransferForm.tsx]: Got it, sounds like fragments are the way to go here, updated: https://github.com/helium/hotspot-app/pull/479/commits/9e2cfb92dc60f3f8dfa082c4641520c6c2fd4cb7?w=1

File: src/features/wallet/send/SendView.tsx (hunk 14 above)
Comment (MichaelBrew) [reply src/features/wallet/send/SendView.tsx]: Ohh very interesting, I did not realize the fee was calculated on a per-transaction basis (that could include 1 to many payments). Does the fee calculation itself have any dependence on the payment amounts? Wondering if the fee calculation should happen at the `SendForm` level instead of on each `SendDetailsForm` level.

For example, if you made 1 payment in a transaction for 5 HNT vs. 100 payments in a transaction for 50 HNT each, would both scenarios incur the same fee?

Comment (MichaelBrew) [reply src/features/wallet/send/SendView.tsx]: Have made the immediate change [here](https://github.com/helium/hotspot-app/pull/479/commits/b2acc92aca47b91b5cfa353265f47570b08757ba), but lmk if it the above comment is still applicable (moving fee calculation up a level)

File: src/features/wallet/send/SendView.tsx [hunk 16]
```diff
@@ -47,14 +39,16 @@ import {
...
+import { decimalSeparator, groupSeparator } from '../../../utils/i18n'
 import { useAppDispatch } from '../../../store/store'
 import {
   fetchCurrentOraclePrice,
   fetchPredictedOraclePrice,
 } from '../../../store/helium/heliumDataSlice'
-import { AppLink, AppLinkType } from '../../../providers/appLinkTypes'
+import {
+  AppLink,
+  AppLinkCategoryType as AppLinkType,
```
MichaelBrew, line 50: I might be missing something with TypeScript, but it doesn't look like `appLinkTypes.ts` exports an `AppLinkType`? My editor also complained about this, but seems like `AppLinkCategoryType` is what we want so I just imported that under a different name.

MichaelBrew:
@tyler-whitman @matthewcarlreetz thanks for the reviews so far, just pushed some more updates based on feedback as well as resolved several merge conflicts (resulting from https://github.com/helium/hotspot-app/pull/518) so would appreciate some eyes on this again. Have also tested with some sample QR codes and seems to be pre-filling as expected, but would love some more help testing!

rawrmaan:
One thing that's important is that every byte in a QR code is precious, so it is likely a good idea to go with a less verbose JSON format in order to more easily support QR codes with up to 50 payees.

### Change by example

#### Old format

```json
{
  "type": "payment",
  "payees": [
    {
      "address": "13JJAmn87e74xR24w6mWKZteHwPaBd1agV6hqJJq22bFFMKVzDe",
      "amount": 23.47292737
    },
    {
      "address": "14L8jfWjDWMVD3Z31ZAEWuj15ZiKZvbYoS14BxF7Tptqvwy129W",
      "amount": 0.67279981
    },
    {
      "address": "14dAR7ad55SV8ELWasZ6aENhrxP77ohiMZEwCg8opmybTkicYCC",
      "amount": 4.54
    }
  ]
}
```

#### New format

```json
{
  "type": "payment",
  "payees": {
    "13JJAmn87e74xR24w6mWKZteHwPaBd1agV6hqJJq22bFFMKVzDe": 23.47292737,
    "14L8jfWjDWMVD3Z31ZAEWuj15ZiKZvbYoS14BxF7Tptqvwy129W": 0.67279981,
    "14dAR7ad55SV8ELWasZ6aENhrxP77ohiMZEwCg8opmybTkicYCC": 4.54
  }
}
```

Of course this only needs to happen at the QR scan level--the object can then be transformed into the old format to prevent changing UI code. Or if the new format is fully adopted, you can still use `Object.keys(payees)` to iterate over the payees in any place that you need to.

tyler-whitman:
> One thing that's important is that every byte in a QR code is precious, so it is likely a good idea to go with a less verbose JSON format in order to more easily support QR codes with up to 50 payees.
> 
> ### Change by example
> #### Old format
> ```json
> {
>   "type": "payment",
>   "payees": [
>     {
>       "address": "13JJAmn87e74xR24w6mWKZteHwPaBd1agV6hqJJq22bFFMKVzDe",
>       "amount": 23.47292737
>     },
>     {
>       "address": "14L8jfWjDWMVD3Z31ZAEWuj15ZiKZvbYoS14BxF7Tptqvwy129W",
>       "amount": 0.67279981
>     },
>     {
>       "address": "14dAR7ad55SV8ELWasZ6aENhrxP77ohiMZEwCg8opmybTkicYCC",
>       "amount": 4.54
>     }
>   ]
> }
> ```
> 
> #### New format
> ```json
> {
>   "type": "payment",
>   "payees": {
>     "13JJAmn87e74xR24w6mWKZteHwPaBd1agV6hqJJq22bFFMKVzDe": 23.47292737,
>     "14L8jfWjDWMVD3Z31ZAEWuj15ZiKZvbYoS14BxF7Tptqvwy129W": 0.67279981,
>     "14dAR7ad55SV8ELWasZ6aENhrxP77ohiMZEwCg8opmybTkicYCC": 4.54
>   }
> }
> ```
> 
> Of course this only needs to happen at the QR scan level--the object can then be transformed into the old format to prevent changing UI code. Or if the new format is fully adopted, you can still use `Object.keys(payees)` to iterate over the payees in any place that you need to.

This is a great point. @MichaelBrew would you be willing to make this change? Or we can always make it another defect and focus on getting your PR merged in its current state.

src/features/wallet/send/SendView.tsx
tyler-whitman: Yeah there was a bug in the code. You are correct. I fixed it in one of my open PRs too, but this will probably merge first so ill just resolve the conflict.

src/features/wallet/send/SendView.tsx
tyler-whitman: Yeah thats a good point, the fee is based on the byte size of the transaction. So adding / removing any data from it will change it. Having more payee / payer pairs could definitely increase the fee so we would want to calculate it with that in mind.

tyler-whitman: I think currently we show the fee under the amount, which isnt really correct when you have multiple payer / payee combos. We may need to move it. Im thinking possibly below the send button so that its always visible and you won't miss it due to having to scroll.

MichaelBrew: Great suggestion - made the change to calculate 1 fee based on their entire transaction and show it under the submit button with https://github.com/helium/hotspot-app/pull/479/commits/d3d08a2aa540b13f46a46668514574dcd6e95846 and https://github.com/helium/hotspot-app/pull/479/commits/092235e5a3402c4c6ad94adf6c8b9c84b4a5d1bd

MichaelBrew:
@rawrmaan @tyler-whitman no problem - if we don't need to support memos in scanned payment codes, can just use that direct payee -> amount mapping. Added here (lmk if there's a more elegant way to handle that amount typing btw): https://github.com/helium/hotspot-app/pull/479/commits/6311549407ef167b60531507edbb18b05d3bf4d3

src/features/wallet/send/SendView.tsx
tyler-whitman: Awesome this looks great

MichaelBrew:
@tyler-whitman is there anything else that needs addressing before we can merge? You'll have probably noticed I only have time to contribute on the weekends, and each week there's a new set of merge conflicts to resolve 😬  would like to get this merged sooner than later to avoid rolling conflicts each week.

Would it help to break this into smaller PRs to mitigate any risk of breaking changes?

tyler-whitman:
> @tyler-whitman is there anything else that needs addressing before we can merge? You'll have probably noticed I only have time to contribute on the weekends, and each week there's a new set of merge conflicts to resolve 😬 would like to get this merged sooner than later to avoid rolling conflicts each week.
> 
> Would it help to break this into smaller PRs to mitigate any risk of breaking changes?

Hey, yeah we will get this merged right after the current release we are working on today. I'll do some more testing today. I can also help with the conflicts / bug fixes to get it in this week. Could you give me push access to your repo?

tyler-whitman:
> @tyler-whitman thanks for the thoughtful review - have pushed several changes based on feedback, including the limitation of multiple "send" actions to only the "payment" send type, and submitting those payments in a single transaction ([ef61ccb](https://github.com/helium/hotspot-app/commit/ef61ccb775622351ce1eaf7d90f3d98e13e70d73)).
> 
> As indicated by the "TBD" label in the PR description, would love some guidance on how to test these changes. This is my first (non-one-liner) PR to the repo so still learning best practices 🙂 Is there any sort of automated test suite that can be run against this branch? Or do we need to manually test? Is there a sandbox/dev/staging environment? Or do things have to be tested against production?

I will update the format to also support memos, we can take a combination of the older version and this current one:

```
{
    "type": "payment",
    "payees": {
        "address": {amount: 1, memo:"asdf"}
    }
}
```

tyler-whitman:
I opened this in another PR (https://github.com/helium/hotspot-app/pull/596) so I could make changes to it and we can get it merged. I fixed the conflicts and a couple bugs that I found but I think we will be able to merge it tomorrow. Thanks again for making this change! Sorry it has taken so long to get it merged.

MichaelBrew:
Awesome, thanks @tyler-whitman! Looking forward to many more contributions 🙂

PR #480, opened 2021-04-19
Handle rewards loading states.

### LINKED ISSUE #409: Graphics glitch
Steps to recreate

• on hotspot home page, hit settings
• when the modal is active, dismiss the modal
• on the nav bar hit 'accounts'

the hotspot > account transition glitches a little

iOS 14.5

matthewcarlreetz:
Fix flash when moving between tabs
Show loading states when rewards not yet loaded
Update hotspot list item style
Fix null,null,null location name
Remove comma after greeting
Closes #409


matthewcarlreetz:
> Looks good and working well. So was it the animateTransitions causing the flickers?

Yeah, it happens when they trigger and the current tab is being transitioned away from.

PR #482, opened 2021-04-19
Set android target and compile sdk to 30.

### LINKED ISSUE #436: Update Android Target API level
Starting August 2021, new apps will be required to target API level 30 (Android 11) and use the Android App Bundle publishing format. Starting November 2021, all app updates will be required to target API level 30 (Android 11).

matthewcarlreetz:
Closes #436

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -4,8 +4,8 @@ buildscript {
     ext {
         buildToolsVersion = "29.0.2"
         minSdkVersion = 21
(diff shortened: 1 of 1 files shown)


matthewcarlreetz:
> Did you test it out with an SDK version 30 phone (Android 11)? We should make sure this doesn't break anything.

I tested it on my emulator, seemed to work fine.

PR #486, opened 2021-04-19
Send Max Fix

### LINKED ISSUE #458: Investigate why Send Max leads to {invalid,insufficient_balance}
Based on the logic we've added in the app when a user sends max, we should not expect to see many failures due to insufficient balance. 

Things to check:

pessimistic oracle price - are we doing this?
when are we calculating transaction fee (double check that it's on send screen, not on app load)
* maybe we need to add a 'recalculate fee' button after 30s of inactivity

if all this checks out, we'll need to see how the consensus group processes these transactions and understand why it's rejecting them.

tyler-whitman:
Ensure we have current oracles and are picking the minimum from the predicted and current prices.

fixes #458

### DIFF
--- a/src/utils/fees.ts
+++ b/src/utils/fees.ts
@@ -27,7 +27,7 @@ export const useFees = () => {
       if (!balance) return new Balance<DataCredits>(0, CurrencyType.dataCredit)
 
       const prices = [currentOraclePrice, ...predictedOraclePrices]
-      const oraclePrice = minBy(prices, (p) => p?.price.integerBalance || 0)
+      const oraclePrice = minBy(prices, (p) => p?.price.integerBalance)
       // ensure precision is only 8 decimals
       const feeHNTInteger = Math.trunc(
         balance.toNetworkTokens(oraclePrice?.price).integerBalance,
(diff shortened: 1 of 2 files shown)


src/utils/fees.ts
```diff
@@ -27,7 +27,7 @@ export const useFees = () => {
       if (!balance) return new Balance<DataCredits>(0, CurrencyType.dataCredit)
 
       const prices = [currentOraclePrice, ...predictedOraclePrices]
```
matthewcarlreetz, line 29: If `currentOraclePrice` is undefined, would this fail?

matthewcarlreetz (approved):

src/utils/fees.ts
tyler-whitman: In the old app it didn't default to 0, I think lodash just ignores the undefined ones. If you default it to 0, then the min would always be 0 and the fee would be incorrect.

matthewcarlreetz: ok, that works.

Entaum:
I am still experiencing this issue. Solution has been to deduct 0.00000001 of the total amount, then it works. Seems like a rounding issue perhaps?

mfalkvidd:
I am also still experiencing this issue. Happened yesterday and today.

PR #492, opened 2021-04-20
Hotspot Details UI Changes

### LINKED ISSUE #327: Swipe left/right for next hotspot card
When trying to view all my hotspot details, I find it slightly painful to have to swipe a hotspot card down, hit Back, then choose another hotspot to get to the next hotspot details. Would be easier if I could swipe left/right to go from one hotspot card to the next.

### LINKED ISSUE #483: Hotspot Details - Overview update
Update how the Hotspot details overview tab look on the app:

This should reduce the white space and compact the data a bit more, and makes it easier to navigate.

In addition to updating this page, we will add back the ability to swipe between hotspots. The next Hotspot in the list will correspond with the filter they have set on the main Hotspot tab.

Top map: show Hotspot location for now (ignore witness dots).
image

matthewcarlreetz:
closes #483 
closes #327 

Move follow flag to top bar.
Update title and status indicators.
Update HotspotsPicker style
Rename BubbleSelect to HeliumSelect and style variants
Use flat variant in place of HeliumActionSheet in hotspots picker
Begin updating style of hotspot detail chart.
Remove ham menu, replace with more and add copy option to share menu.
Add prev/next hotspot buttons to hotspot details.

https://user-images.githubusercontent.com/1760821/115451055-5115dc00-a1e2-11eb-9cc3-46a5a1f8bcc3.mov


src/features/hotspots/details/HexBadge.tsx
```diff
@@ -28,17 +30,17 @@ const HexBadge = ({ rewardScale }: Props) => {
...
-      paddingStart="s"
-      height="100%"
-      marginLeft="s"
+      width={59}
+      height={30}
+      marginLeft="xs"
     >
-      <Hex color="white" width={14} />
-      <Text color="white" variant="body2Medium" paddingHorizontal="s">
+      <Hex color={greenOnline} width={14} />
```
tyler-whitman, line 42: Should we do different colors here depending on the ratio?

src/features/hotspots/details/HotspotDetailsHandle.tsx
```diff
@@ -1,28 +1,41 @@
...
       alignItems="center"
     >
-      <FollowButton
-        paddingHorizontal="l"
-        paddingVertical="m"
-        address={hotspot.address}
-      />
-      <CardHandle />
-      <HotspotMoreMenuButton hotspot={hotspot} />
+      <TouchableOpacityBox onPress={prev} padding="m" style={styles.prevButton}>
```
tyler-whitman, line 27: Maybe we should debounce these?

src/features/hotspots/details/HotspotDetailsHandle.tsx
```diff
@@ -1,28 +1,41 @@
...
       flex={1}
       justifyContent="space-between"
       alignItems="center"
     >
-      <FollowButton
-        paddingHorizontal="l"
-        paddingVertical="m"
-        address={hotspot.address}
-      />
-      <CardHandle />
```
tyler-whitman, line 22: Do we still want the handle?

src/features/hotspots/details/HotspotDetailsHandle.tsx
```diff
@@ -1,28 +1,41 @@
...
-      <HotspotMoreMenuButton hotspot={hotspot} />
+      <TouchableOpacityBox onPress={prev} padding="m" style={styles.prevButton}>
+        <Chevron color="#C2C5E4" height={20} width={20} />
+      </TouchableOpacityBox>
+      <TouchableOpacityBox onPress={next} padding="m">
```
tyler-whitman, line 30: Should we remove the arrows if you haven't selected on your own or followed hotspots?

src/features/hotspots/details/HotspotDetailsHandle.tsx
```diff
@@ -1,28 +1,41 @@
...
+
+const HotspotDetailsHandle = () => {
+  const dispatch = useAppDispatch()
+
+  const prev = useCallback(() => {
+    dispatch(hotspotsSlice.actions.selectPrevHotspot())
+  }, [dispatch])
+
+  const next = useCallback(() => {
+    dispatch(hotspotsSlice.actions.selectNextHotspot())
```
tyler-whitman, line 17: If you select the followed hotspots filter when you have no followed hotspots and then go to the map and select a hotspot and tap the next arrow, you end up on a blank white card

tyler-whitman: Also if you are following only one hotspot or I'm guessing have only one hotspot, the next and previous don't do anything. Maybe we should hide them?

tyler-whitman (changes requested):

src/features/hotspots/details/HexBadge.tsx
tyler-whitman: No, but explorer does https://github.com/helium/explorer/blob/master/components/Hotspots/utils.js#L101

src/features/hotspots/details/HotspotDetailsHandle.tsx
tyler-whitman: Ok sounds good

src/features/hotspots/details/HexBadge.tsx
matthewcarlreetz: Nice, I'll check it out.

src/features/hotspots/details/HotspotDetails.tsx
```diff
@@ -55,6 +62,12 @@ const HotspotDetails = ({ hotspot }: { hotspot?: Hotspot }) => {
 
   const [timelineValue, setTimelineValue] = useState(14)
 
+  const syncStatus = useMemo(() => {
+    if (!hotspot?.block) return
+
+    return getSyncStatus(hotspot.block, blockHeight)
```
tyler-whitman, line 68: I forget what this block is, but we should be using hotspot.status.height instead. Also I think we should only show "Needs Attention" if the hotspot.status.online does not equal "online" to match explorer. The status data is the gossiped data that we use for online state and is also what controls the offline / online notifications so we need to match that.

src/features/hotspots/details/HotspotDetails.tsx
```diff
@@ -55,6 +62,12 @@ const HotspotDetails = ({ hotspot }: { hotspot?: Hotspot }) => {
 
   const [timelineValue, setTimelineValue] = useState(14)
 
+  const syncStatus = useMemo(() => {
+    if (!hotspot?.block) return
```
tyler-whitman, line 66: We shouldnt need this check when we use `hotspot.status`

src/features/hotspots/details/HotspotStatusBanner.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const title = useMemo(() => {
+    if (hotspot.status?.online === 'online') {
+      return t('hotspot_details.status_prompt_online.title')
+    }
+    return t('hotspot_details.status_prompt_offline.title')
+  }, [hotspot.status, t])
+
+  const subtitle = useMemo(() => {
+    if (hotspot.status?.online !== 'online') return
```
tyler-whitman, line 55: I am seeing the banner because the check I mentioned above is incorrect, and then this case is returning show the block data is missing. Once you move it to use `hotspot.status` in the other place this should be fixed

tyler-whitman (changes requested):

tyler-whitman (approved):

PR #497, opened 2021-04-21
Add antenna configuration for LongAP One

timcooijmans:


src/components/HotspotConfigurationPicker.tsx
```diff
@@ -38,6 +39,7 @@ export const Antennas: Record<AntennaId, Antenna> = {
   syncrobit_us: { id: 'syncrobit_us', gain: 1.2 },
   syncrobit_eu: { id: 'syncrobit_eu', gain: 2.3 },
   rak_custom: { id: 'rak_custom', gain: 5.8 },
+  longapone_eu: {id: 'longapone_eu', gain: 3 },
```
tyler-whitman, line 42: Can you add a space before id `longapone_eu: { id: 'longapone_eu', gain: 3 },`

tyler-whitman (changes requested):
Also in https://github.com/helium/hotspot-app/blob/main/src/features/hotspots/setup/AntennaSetupScreen.tsx#L34 you will need to add a case for LONGAPONE

```
      case 'LONGAPONE':
        return Antennas.longapone_eu
```

timcooijmans:
All comments fixed!

tyler-whitman (approved):

PR #500, opened 2021-04-21
Center hotspot setup content vertically.

### LINKED ISSUE #306: Content positioning on some screens
I don't know if these are absolutely positioned in all cases, or if we're able to 'flex box' it, so that all content is roughly vertically centred on the screen.

Screenshot 2021-02-25 at 09 05 52

Also noticed on 'Power up' screen, but could be others (some of these were initially going to have images, which we wont do now).

matthewcarlreetz:
closes #306

### DIFF
--- a/src/features/hotspots/setup/HotspotSetupBluetoothError.tsx
+++ b/src/features/hotspots/setup/HotspotSetupBluetoothError.tsx
@@ -20,13 +20,13 @@ const HotspotSetupBluetoothError = () => {
   const navigation = useNavigation<HotspotSetupNavigationProp>()
   return (
     <Box flex={1}>
-      <Box flex={1}>
-        <Box marginBottom="l">
-          <Bluetooth />
-        </Box>
+      <Box flex={1} alignItems="center" justifyContent="center">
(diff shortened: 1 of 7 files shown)


src/features/hotspots/setup/HotspotSetupBluetoothError.tsx
```diff
@@ -19,68 +19,63 @@ const HotspotSetupBluetoothError = () => {
...
+            backgroundColor="purpleMain"
+            shadowColor="purpleMain"
+            shadowOpacity={1}
+            shadowOffset={{ width: 0, height: 0 }}
+            shadowRadius={8}
+          />
         </Box>
       </Box>
 
-      <Box justifyContent="flex-end">
```
tyler-whitman, line 74: The buttons are no longer positioned at the bottom, which doesn't follow the mock up. 

For example the diagnostic screen while onboarding still had a bottom fixed button but centered text content.

tyler-whitman (changes requested):

tyler-whitman (approved):

PR #503, opened 2021-04-21
Remove one signal prompt on android.

matthewcarlreetz:


tyler-whitman (approved):
Do we not need this call on Android?

matthewcarlreetz:
> Do we not need this call on Android?

We don't. Before this you'd always seen a warning on android 
> promptForPushNotificationsWithUserResponse: this function is not supported on Android

PR #505, opened 2021-04-22
open notification list when tapping push notifications

### LINKED ISSUE #442: Push Notification Deep Links
We should support deep linking into the app from push notifications. Right now when you tap a notification it only opens the app to the default screen.

For example if you tap a transfer notification it should open the notification tab so you see the new transfer request.

tyler-whitman:
closes #442 

This makes it so that when you tap a push notification the app opens to the notifications screen after unlocking. This way when users get a notification and tap it they will see the new notification after opening the app.

### DIFF
--- a/src/features/hotspots/root/HotspotsTicker.tsx
+++ b/src/features/hotspots/root/HotspotsTicker.tsx
@@ -61,11 +61,16 @@ const HotspotsTicker = ({ ...boxProps }: Props) => {
     const formattedBlockTime = blockTime?.toLocaleString(locale, {
       maximumFractionDigits: 0,
     })
-    return t('hotspots.ticker', {
-      formattedHotspotCount,
-      oraclePrice,
-      formattedBlockTime,
-    })
+    return blockTime
+      ? t('hotspots.ticker', {
+          formattedHotspotCount,
+          oraclePrice,
+          formattedBlockTime,
+        })
+      : t('hotspots.ticker_no_block', {
+          formattedHotspotCount,
+          oraclePrice,
+        })
   }, [blockTime, currentOraclePrice?.price, hotspotCount, t])
 
   return (
--- a/src/features/hotspots/root/HotspotsTicker.tsx
+++ b/src/features/hotspots/root/HotspotsTicker.tsx
@@ -58,7 +58,7 @@ const HotspotsTicker = ({ ...boxProps }: Props) => {
   const text = useMemo(() => {
     const formattedHotspotCount = hotspotCount?.toLocaleString(locale) || 0
     const oraclePrice = currentOraclePrice?.price || 0
-    const formattedBlockTime = blockTime.toLocaleString(locale, {
+    const formattedBlockTime = blockTime?.toLocaleString(locale, {
       maximumFractionDigits: 0,
     })
     return t('hotspots.ticker', {
--- a/src/navigation/main/MainTabNavigator.tsx
+++ b/src/navigation/main/MainTabNavigator.tsx
@@ -4,7 +4,7 @@ import { useNavigation } from '@react-navigation/native'
 import { useSelector } from 'react-redux'
 import { StyleSheet } from 'react-native'
 import Hotspots from '../../features/hotspots/root/HotspotsNavigator'
-import { TabBarIconType, MainTabType, RootNavigationProp } from './tabTypes'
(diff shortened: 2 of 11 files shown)


src/store/notifications/notificationSlice.ts
```diff
@@ -0,0 +1,24 @@
+import { createSlice } from '@reduxjs/toolkit'
+import { OSNotification } from 'react-native-onesignal'
+
+export type NotificationState = {
+  notification?: OSNotification
```
matthewcarlreetz, line 5: I think we should move the notifications in `src/store/account/accountSlice.ts` over to here.

matthewcarlreetz (changes requested):

src/features/hotspots/root/HotspotsTicker.tsx
```diff
@@ -58,7 +58,7 @@ const HotspotsTicker = ({ ...boxProps }: Props) => {
   const text = useMemo(() => {
     const formattedHotspotCount = hotspotCount?.toLocaleString(locale) || 0
     const oraclePrice = currentOraclePrice?.price || 0
-    const formattedBlockTime = blockTime.toLocaleString(locale, {
+    const formattedBlockTime = blockTime?.toLocaleString(locale, {
```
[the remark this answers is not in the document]
tyler-whitman: Ok let me see, I was fixing some state bug here with the types

src/navigation/main/MainTabNavigator.tsx
```diff
@@ -14,29 +14,41 @@ import WalletNavigator from '../../features/wallet/WalletNavigator'
...
 import appSlice from '../../store/user/appSlice'
 import NotificationsScreen from '../../features/notifications/NotificationsScreen'
+import notificationSlice from '../../store/notifications/notificationSlice'
 
 const MainTab = createBottomTabNavigator()
 
 const MainTabs = () => {
   const { white, grayLight } = useColors()
-  const navigation = useNavigation<RootNavigationProp>()
+  const navigation = useNavigation()
```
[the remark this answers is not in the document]
tyler-whitman: Ok I was trying to figure out what to do there, is there any problem with creating two navigators on one screen?

src/store/notifications/notificationSlice.ts
tyler-whitman: Ah ok I can, maybe I call this new one pushNotifcation?

src/navigation/main/MainTabNavigator.tsx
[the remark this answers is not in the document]
matthewcarlreetz: It should be ok. MainTabNav is nested inside of RootNat.

src/store/notifications/notificationSlice.ts
matthewcarlreetz: that sounds good.

src/navigation/main/MainTabNavigator.tsx
```diff
@@ -20,34 +25,35 @@ const MainTab = createBottomTabNavigator()
...
-    (state: RootState) => state.notifications.notification,
+  const pushNotification = useSelector(
+    (state: RootState) => state.notifications.pushNotification,
   )
   const dispatch = useAppDispatch()
 
   useEffect(() => {
     if (!isLocked) return
-    navigation.navigate('LockScreen', { requestType: 'unlock', lock: true })
+    navigation.push('LockScreen', { requestType: 'unlock', lock: true })
```
matthewcarlreetz: Just curious why these were changed from navigate to push?

matthewcarlreetz (approved):
💯

src/navigation/main/MainTabNavigator.tsx
tyler-whitman: Thats what they were originally, I was actually unsure on the difference between the two but I decided to put them back to what they were previously.

tyler-whitman: Ah looks like the only difference is push allows the component to be pushed on the stack multiple times. In the case they have the same effect

allenan (approved):
nice this looks good, the one thing i mentioned was to see if we can trigger the new notifications circle icon on the nav bar to show up when a push notification comes through. definitely like that we're not polling notifications based on blocks anymore 👍

PR #509, opened 2021-04-23
hotspot detail app level caching

### LINKED ISSUE #490: App Level caching
cache data when users tap between tabs to further ease burden on API.

tyler-whitman:
closes #490 

With this change we cache the hotspot detail data for 5 minutes.

### DIFF
--- a/src/store/hotspotDetails/hotspotDetailsSlice.ts
+++ b/src/store/hotspotDetails/hotspotDetailsSlice.ts
@@ -114,7 +114,7 @@ export const fetchHotspotData = createAsyncThunk<HotspotData, string>(
   async (address: string, { getState }) => {
     const currentState = (getState() as {
       hotspotDetails: {
-        hotspotData: Record<string, HotspotDetailCache>
+        hotspotData: HotspotIndexed<HotspotDetailCache>
       }
     }).hotspotDetails
     const hotspotData = currentState.hotspotData[address] || {}
@@ -140,7 +140,7 @@ export const fetchHotspotChartData = createAsyncThunk<
   async (params: FetchDetailsParams, { getState }) => {
     const currentState = (getState() as {
       hotspotDetails: {
-        chartData: Record<string, Record<number, HotspotChartCache>>
+        chartData: HotspotIndexed<HotspotChartRecord>
       }
     }).hotspotDetails
     const chartData = currentState.chartData[params.address] || {}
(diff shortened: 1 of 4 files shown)


matthewcarlreetz (approved):

src/utils/cacheUtils.ts
```diff
@@ -0,0 +1,36 @@
+export type CacheRecord<T> = T & {
+  lastFetchedTimestamp: number
+  loading: boolean
+}
+
+const getCacheRecord = <T>(item: T) => {
```
allenan, line 6: i'm a little confused about this function. based on the naming i was expecting it to do the actual "getting" from the cache, but it looks like it just formats something as a `CacheRecord` and the lookup is done elsewhere?

src/store/hotspotDetails/hotspotDetailsSlice.ts
```diff
@@ -102,9 +109,22 @@ export const fetchHotspotChallengeSums = async (
...
+export const fetchHotspotDetails = createAsyncThunk<
+  HotspotDetailsObj,
+  FetchDetailsParams
+>(
   'hotspotDetails/fetchHotspotDetails',
-  async (params: FetchDetailsParams) => {
+  async (params: FetchDetailsParams, { getState }) => {
+    const currentState = (getState() as {
+      hotspotDetails: {
+        details: Record<string, Record<number, HotspotDetails>>
```
allenan, line 120: is this type here: `Record<string, Record<number, HotspotDetails>>` a cache record? is it the same as `CacheRecord<HotspotDetails>`? seems like it could be a more generic/reusable type

src/utils/cacheUtils.ts
tyler-whitman: Yeah it just does conversion and verifies the data is of the correct type. Maybe we can call it `convertToCacheRecord()`?

src/store/hotspotDetails/hotspotDetailsSlice.ts
tyler-whitman: No, the `CacheRecord` only adds loading and a lastFetchedTimestamp to the data. So any object can become a cache record and get those extra properties to manage its caching lifecycle. Maybe its more clear now that I did the refactoring?

src/utils/cacheUtils.ts
allenan: or like `makeCacheRecord` or `toCacheRecord`? `get` just comes with some overloaded assumptions

tyler-whitman: Yeah, what about `asCacheRecord` since its mostly typing

PR #513, opened 2021-04-26
Use pressable to fix HotspotListItem pressed background color.

matthewcarlreetz:


src/components/HotspotListItem.tsx
```diff
@@ -70,79 +70,82 @@ const HotspotListItem = ({
...
-              <Text
-                variant="body2Medium"
-                color="offblack"
-                paddingStart="s"
-                ellipsizeMode="tail"
-                numberOfLines={2}
-                maxWidth={210}
-              >
-                {animalName(hotspot.address)}
+      <Pressable onPress={handlePress}>
```
tyler-whitman, line 73: Can't we just use a TouchableHighlight?

matthewcarlreetz: We definitely could, but according to the docs `Pressable` is the way of the future, so wanted to try it out.

TouchableHighlight · React Native 2021-04-26 at 12 32 34 PM

tyler-whitman: Ok haha sounds good

tyler-whitman (approved):

PR #516, opened 2021-04-27
show loading indicator for initial notification fetch

### LINKED ISSUE #515: Notifications Loading Indicator
The notifications screen has been changed to query on open or when landing on the screen, so it can take a few seconds before everything renders and appears.

Add a loading indicator to show that things are happening.

tyler-whitman:
closes #515

### DIFF
--- a/src/features/notifications/NotificationsScreen.tsx
+++ b/src/features/notifications/NotificationsScreen.tsx
@@ -1,5 +1,6 @@
 import React, { useCallback } from 'react'
 import { useSelector } from 'react-redux'
+import { ActivityIndicator } from 'react-native'
 import SafeAreaBox from '../../components/SafeAreaBox'
 import { RootState } from '../../store/rootReducer'
 import { useAppDispatch } from '../../store/store'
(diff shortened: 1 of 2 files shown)


src/features/notifications/NotificationsScreen.tsx
```diff
@@ -44,7 +44,9 @@ const NotificationsScreen = () => {
           refreshing={markNotificationStatus === 'pending'}
         />
       )}
-      {notifications.length === 0 && <EmptyNotifications />}
+      {notifications.length === 0 && markNotificationStatus !== 'pending' && (
```
matthewcarlreetz, line 47: This will keep the `EmptyNotifications` view from showing, but won't show an `ActivityIndicator`, right?

matthewcarlreetz (approved):

src/features/notifications/NotificationsScreen.tsx
tyler-whitman: Oh good point need one for the loading above too so it shows refreshing

PR #534, opened 2021-05-04
payment memos

### LINKED ISSUE #461: Display memos in payment_v2
Blockchain team is allowing memos in payment_v2 transactions. https://github.com/helium/blockchain-core/pull/768

To start, we should support and show the `Memo` field in transaction details for send and receive transaction screens.

image

tyler-whitman:
Needs changes from https://github.com/helium/helium-js/pull/167 before it can merge.

Adds payment v2 memos.

The flow will be:

1. User enters up to 8 bytes of text in the payment memo field
2. We encode that text from utf8 to base64 and send it to helium-js. Helium-js encodes the base 64 to a long representation fo the bytes.
3. The API sends us the base64 string and we can decode it back to utf8 to get the plain text.

closes #461


src/features/wallet/root/ActivityDetails/Payment.tsx
```diff
@@ -21,14 +22,23 @@ const Payment = ({ item, address }: Props) => {
...
       {payments.map((p, index) => (
-        <PaymentItem
-          key={p.payee}
-          text={p.payee}
-          isMyAccount={p.payee === address}
-          mode="to"
-          isFirst={false}
-          isLast={index === payments.length - 1}
-        />
+        <Box key={p.payee}>
```
matthewcarlreetz, line 25: Will `payee` always be unique?

src/locales/en.ts
```diff
@@ -515,6 +515,10 @@ export default {
     memo: {
       label: 'Memo',
       placeholder: 'Enter Memo... (optional)',
+      bytes_left: '{{count}} byte left',
+      bytes_left_plural: '{{count}} bytes left',
+      length_error:
+        'Memo is too long. Please edit the memo to 8 bytes or less.',
```
matthewcarlreetz, line 521: Do you think "bytes" is a concept users will understand?

src/utils/fees.ts
```diff
@@ -67,6 +68,7 @@ export const calculatePaymentTxnFee = async (
       {
         payee,
         amount,
+        memo: memo || undefined,
```
matthewcarlreetz, line 71: Can this just be 
```
    payments: [
      {
        payee,
        amount,
        memo,
      },
    ],
```
?

matthewcarlreetz (approved):
Looks good. Just a couple questions for you.

src/locales/en.ts
tyler-whitman: Good question. So it shows the number of bytes you are using as you type. The reason we cant say "8 characters" or something is because emojis or some other language characters are more than 1 byte.

I think with it showing how many bytes you are using as you type you kind of get it even if you don't know what bytes are. Any idea on a way we could make it better?

src/features/wallet/root/ActivityDetails/Payment.tsx
tyler-whitman: I think so, but let me double check you cant have the same payee multiple times.

src/utils/fees.ts
tyler-whitman: Actually it needed to be `encodeMemoString(memo)` good catch haha

src/features/wallet/root/ActivityDetails/Payment.tsx
tyler-whitman: The blockchain doesn't allow it https://github.com/helium/blockchain-core/blob/master/test/blockchain_payment_v2_SUITE.erl#L196-L219

PR #547, opened 2021-05-06
Allow user to retry when onboarding key is null.

matthewcarlreetz:


src/utils/useHotspot.ts
```diff
@@ -32,6 +32,14 @@ import { RootState } from '../store/rootReducer'
 import * as Logger from './logger'
 import useSubmitTxn from '../hooks/useSubmitTxn'
 
+export type HotspotConnectStatus =
```
tyler-whitman, line 35: Nice

src/utils/useHotspot.ts
```diff
@@ -111,14 +119,22 @@ const useHotspot = () => {
...
+
+  const connectAndConfigHotspot = async (
+    hotspotDevice: Device,
+  ): Promise<HotspotConnectStatus> => {
     let connectedDevice = hotspotDevice
     const connected = await hotspotDevice.isConnected()
     if (!connected) {
       const device = await connect(hotspotDevice)
       if (!device) {
-        Logger.error(new Error('Hotspot connect failed: no device'))
```
tyler-whitman, line 120: Can we leave the logs in, just so we can tell when things are going wrong and where for people reporting issues

tyler-whitman (approved):

src/utils/useHotspot.ts
matthewcarlreetz: handleConnectStatus will log

PR #551, opened 2021-05-06
Use phone location when disco mode on unasserted hotspot.

### LINKED ISSUE #540: Allow disco mode for unasserted Hotspots
It'd be great if users can initiate discovery mode on an unasserted Hotspot, as a common use case is identifying ideal hotspot placement before setting a location and using up their free assert.

Without a hotspot location, we won't know what to show on the map. If we assume that the phone's location is where the Hotspot is, we can request location permissions and use the phone as proxy for the Hotspot and continue with disco mode.

User taps Disco Mode on an unasserted Hotspot > Begin Session
Show popup:
"Hotspot Does Not Have A Location"
"To visualize Hotspots that respond, we will use your phone's location as a placeholder for the Hotspot."
"Cancel | Continue"

When they hit Continue, request location permissions if we don't have them. If we have their location permission, continue with DM session and use their phone's location as the central point in the map.

matthewcarlreetz:
closes #540

### DIFF
--- a/src/features/hotspots/settings/discovery/DiscoveryMap.tsx
+++ b/src/features/hotspots/settings/discovery/DiscoveryMap.tsx
@@ -35,7 +35,7 @@ export type MapSelectDetail = {
   address: string
 }
 type Props = BoxProps<Theme> & {
-  hotspotCoords: number[]
+  coords: number[]
   hotspotAddress: string
   responses: DiscoveryResponse[]
   onSelect: ({ lat, lng, name }: MapSelectDetail) => void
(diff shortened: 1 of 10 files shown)


src/features/hotspots/settings/discovery/DiscoveryMap.tsx
```diff
@@ -217,7 +217,7 @@ const DiscoveryMap = ({
         zoomEnabled={false}
         onDidFinishLoadingMap={setupMap}
       >
-        <MapboxGL.Camera ref={cameraRef} />
+        <MapboxGL.Camera ref={cameraRef} zoomLevel={12} />
```
tyler-whitman, line 220: There is also a `defaultSettings` zoomLevel, I think in Map.tsx I needed that to get it to load at the correct zoom level. Maybe this works though if it doesnt change?

tyler-whitman (approved):

PR #555, opened 2021-05-10
Jank Fixes

matthewcarlreetz:
Fix jank when initially loading hotspot list
Fix jank when foregrounding app from activity tab then returning to hotspots


src/features/hotspots/root/HotspotsView.tsx
```diff
@@ -129,6 +129,14 @@ const HotspotsView = ({
     return [1, listHeight.current]
   }, [detailHeaderHeight, listHeight, showDetails])
 
+  const handleMapLoaded = useCallback(() => {
```
matthewcarlreetz, line 132: The map is a beast and causes jankiness - wait for it to load to reveal the list.

ios/Helium.xcodeproj/xcshareddata/xcschemes/Helium.xcscheme
```diff
@@ -41,9 +41,9 @@
       </Testables>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Debug"
-      selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
-      selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      buildConfiguration = "Release"
```
tyler-whitman, line 44: Should we keep this debug?

src/features/hotspots/root/HotspotsView.tsx
```diff
@@ -124,6 +129,17 @@ const HotspotsView = ({
...
   }, [params])
 
+  useMount(() => {
+    if (startOnMap) {
+      setShowMap(true)
+      return
+    }
+
+    setTimeout(() => {
+      setShowMap(true)
```
tyler-whitman, line 139: So the map loads after the sheet now?

tyler-whitman (approved):

ios/Helium.xcodeproj/xcshareddata/xcschemes/Helium.xcscheme
matthewcarlreetz: yeah, committed by mistake

src/features/hotspots/root/HotspotsView.tsx
matthewcarlreetz: yeah, it seems better to me this way. It removes the awkward chop that happens when they load at the same time. Could you try it and see what you think? Probably best to do with a release build.

matthewcarlreetz: ok, sweet.

PR #559, opened 2021-05-11
Fix bug when paying/transfer/burning and isPinRequiredForPayment is true

### LINKED ISSUE #558: Unable to transfer Hotspot if "Require PIN for Payment" is set to ON
Reported in Discord and able to reproduce - 
https://discord.com/channels/404106811252408320/730245373037379647/841693455025438770

Set Require PIN for Payments to On, then go to transfer a Hotspot.

Scan QR code to prepopulate the address
See that "You don't have enough HNT", the Hotspot name is missing in the blue bar, and the address is empty, despite scanning the QR code.

IMG_0938D8D8D277-1

app version 3.1.2

matthewcarlreetz:
closes #558


tyler-whitman (approved):
Working well for me now, there is a little delay for the lock screen to pop up but I think its ok.

PR #589, opened 2021-05-18
Add support for stake activity

matthewcarlreetz:


tyler-whitman (approved):
Looks good, I guess for the open TODOs we will have to wait until staking is out, or is there someone we can confirm with?

PR #592, opened 2021-05-19
Move appstate from slice and to appstate lib hook

matthewcarlreetz:
There were some edge cases where the `useVisible` hook was wrong on Android. Hoping this will clear them up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "react-async-hook": "^3.6.2",
     "react-i18next": "^11.7.3",
     "react-native": "0.64.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -93,6 +93,7 @@
     "react-async-hook": "^3.6.2",
     "react-i18next": "^11.7.3",
     "react-native": "0.64.0",
+    "react-native-appstate-hook": "^1.0.5",
```
tyler-whitman, line 96: Nice this makes it easy

tyler-whitman (approved):
Looks good

PR #596, opened 2021-05-20
qr code scan multiple payments

### LINKED ISSUE #134: Allow QR code scanning of payment_v2 transactions

tyler-whitman:
re-opening https://github.com/helium/hotspot-app/pull/479

closes #134


matthewcarlreetz (approved):

tyler-whitman:
@MichaelBrew we are merging this in. It is planned for the next app release v3.2.2

PR #607, opened 2021-05-24
Bugfix/onboarding errors

matthewcarlreetz:


src/utils/assertLocationUtils.tsx
```diff
@@ -23,7 +24,13 @@ export const assertLocationTxn = async (
     return undefined
   }
 
-  const { speculativeNonce = 0 } = await getHotspotDetails(gateway)
+  let speculativeNonce = 0
+  try {
+    const response = await getHotspotDetails(gateway)
```
matthewcarlreetz, line 29: I'm pretty sure this call to `getHotspotDetails` was causing location assertion to fail. It throws a 404 when onboarding a new hotspot. Wrapping it in a try/catch should fix that.

tyler-whitman: Interesting, yeah there was a user complaining about this error. Wonder why it wasn't popping up before but this change looks good.

tyler-whitman (approved):

PR #609, opened 2021-05-25
Fix disco timer when app put in background.

### LINKED ISSUE #603: Disco mode session time paused in app when it goes to sleep
when the phone goes to sleep during an active disco session, the timer pauses, leading users to believe a session can last longer than 5 minutes.

the timer picks up again when the app is in the foreground - this isn't indicative of what happens on the wallet-api side.

matthewcarlreetz:
Closes #603

### DIFF
--- a/src/features/hotspots/settings/discovery/useDiscoveryPoll.ts
+++ b/src/features/hotspots/settings/discovery/useDiscoveryPoll.ts
@@ -1,0 +1,115 @@
+import { Hotspot } from '@helium/http'
+import { getUnixTime } from 'date-fns'
+import { isEqual } from 'lodash'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { useSelector } from 'react-redux'
+import { fetchDiscoveryById } from '../../../../store/discovery/discoverySlice'
+import { ViewState } from '../../../../store/discovery/discoveryTypes'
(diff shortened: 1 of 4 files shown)


src/features/hotspots/settings/discovery/useDiscoveryPoll.ts
```diff
@@ -0,0 +1,115 @@
...
+
+type Props = { viewState: ViewState; hotspot: Hotspot }
+
+const unixTime = (dateStr?: string) => {
+  const insertDate = dateStr ? new Date(dateStr) : null
+  if (!insertDate) return 0
+
+  return getUnixTime(insertDate)
+}
+const useDiscoveryPoll = ({ viewState, hotspot }: Props) => {
```
tyler-whitman, line 19: I have used https://github.com/ocetnik/react-native-background-timer in the past too for background timers. But looks like you got it working without needing it

tyler-whitman (approved):
Looks good, we could use the background timer library or leave as.

src/features/hotspots/settings/discovery/useDiscoveryPoll.ts
matthewcarlreetz: Nice. I'll keep that in mind for future stuff.

PR #610, opened 2021-05-25
handle consecutive pending assert txns

### LINKED ISSUE #605: Handle consecutive pending assert_location transactions better
When a user submits an assert location transaction and they don't see the pending txn UI, they may think that it was not accepted or "didn't go through" and proceed to submit a second assert minutes later.

What happens then, is the consensus group will reject the second assert, user sees a failed notification alert, and the first assert goes through ten+ minutes after. This is confusing.

This can happen when the network is slow or if the user is impatient and submits multiple txns in a row.

-----

We can be better at informing status and preventing users from submitting multiple transactions.

1) after the remote assert on a hotspot, instead of going to hotspots settings, go to the account tab so users can see the pending assert transaction

2) if there is a pending transaction, do not let users tap "Update Hotspot". 
Disable the Update Hotspot button and popup a message 

"Pending Update"
"There is a pending assert location transaction. Please wait for the transaction to clear before submitting another update."
"OK"

tyler-whitman:
closes #605

### DIFF
--- a/src/features/hotspots/settings/HotspotSettings.tsx
+++ b/src/features/hotspots/settings/HotspotSettings.tsx
@@ -156,7 +156,7 @@ const HotspotSettings = ({ hotspot }: Props) => {
     // Check for pending assert
     const pendingTxns = await getAccountTxnsList('pending')
     const txns = (await pendingTxns?.take(20)) as PendingTransaction[]
-    const hasPending = txns?.some(
+    const hasPending = txns?.find(
       (tnx: PendingTransaction) =>
         tnx.txn.type === 'assert_location_v2' &&
         tnx.status === 'pending' &&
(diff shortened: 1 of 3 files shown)


src/features/hotspots/settings/HotspotSettings.tsx
```diff
@@ -150,9 +152,22 @@ const HotspotSettings = ({ hotspot }: Props) => {
...
   }, [setNextState])
 
-  const onPressUpdateHotspot = useCallback(() => {
-    setNextState('updateHotspot')
-  }, [setNextState])
+  const onPressUpdateHotspot = useCallback(async () => {
+    // Check for pending assert
+    const pendingTxns = await getAccountTxnsList('pending')
+    const txns = (await pendingTxns?.take(20)) as PendingTransaction[]
+    const hasPending = txns?.some(
```
matthewcarlreetz, line 159: Little nitpick... I think `pendingTxns?.find()` would be a better fit because it would short circuit when it finds the first one.

matthewcarlreetz (approved):
Seems like this could all pipe through redux, but I'm good with it as is.

PR #617, opened 2021-05-27
Feature/498 qr add hotspot

### LINKED ISSUE #498: Support adding hotspot using (dynamic) QR-code
Can we add a way to replicate the add_gateway behaviour over BLE with a "dynamic” QR-code to the app? So:
1. Arguments for gateway-signed transaction come from some source (manual-input of owner-address + some values, think of management-dashboard either remote or on the hotspot itself)
2. Gateway(= miner currently) creates gateway-signed transaction (just like over BLE)
3. Transaction is presented as QR-Code to app
4. QR-code is scanned by app
5. Details are confirmed by user in app
6. Transaction is signed by app

matthewcarlreetz:
Closes #498 

TODO: Need to add QR hotspot models and messaging specific to those devices.

Can scan the qr by selecting "QR Hotspot Maker" in the hotspot type list or by using the system camera.

Here's the code I've been working with...
add_gateway


src/features/hotspots/setup/HotspotSetupQrConfirmScreen.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  const handleClose = useCallback(() => rootNav.navigate('MainTabs'), [rootNav])
+
+  useEffect(() => {
+    if (!publicKey) return
+
+    const getMac = async () => {
+      const record = await getOnboardingRecord(publicKey)
+      animateTransition('HotspotSetupQrConfirmScreen.GetMac')
+      setMacAddress(record.macEth0)
```
matthewcarlreetz, line 44: Is this the right mac?

timcooijmans: In our case it, is.

ChunChiehKung:
Hi Sir,
Is there a description about the QR code format?
If I want to add a gateway, what information should I provide?
I think  I just need to transfer the information to base64 encoded and create a QR code.
Please advise. Thanks.

package.json
```diff
@@ -68,6 +68,7 @@
     "compare-versions": "^3.6.0",
     "date-fns": "^2.16.1",
     "expo-barcode-scanner": "^9.1.0",
+    "expo-camera": "^11.0.3",
```
tyler-whitman, line 71: We need the camera module now too?

src/providers/AppLinkProvider.tsx
```diff
@@ -76,28 +80,51 @@ const useAppLink = () => {
         case 'transfer':
           navigator.send({ scanResult: record })
           break
+
+        case 'add_gateway': {
+          if (!qrOnboardEnabled) {
+            if (qrOnboardEnabled === undefined) {
```
tyler-whitman, line 86: This check is a little confusing with the one above, can they be combined? When do we expect`qrOnboardEnabled` to be undefined, this is the feature flag right?

src/store/features/featuresSlice.ts
```diff
@@ -28,6 +29,7 @@ const featuresSlice = createSlice({
       state.discoveryEnabled = payload.discoveryEnabled
       state.checklistEnabled = payload.checklistEnabled
       state.followHotspotEnabled = payload.followHotspotEnabled
+      state.qrOnboardEnabled = payload.qrOnboardEnabled
```
tyler-whitman, line 32: Should this be false by default? Did we reuse the list from the last test group?

tyler-whitman (approved):
Looks good, just had a couple questions.

package.json
matthewcarlreetz: Yeah, `expo-barcode-scanner` can't do square frames on Android. They've built scanning into the main camera package now. Seems like they'll probably deprecate the other at some point.

src/store/features/featuresSlice.ts
matthewcarlreetz: It's a different test group. We want it to be undefined at first so that if someone scans a QR code from the system camera we know to set the link to `unhandled` until the API responds with the actual state of `qrOnboardEnabled`.

src/providers/AppLinkProvider.tsx
matthewcarlreetz: It will be `undefined` when the qr is scanned from the system camera and the app is closed. We need to wait for the API to respond before deciding to handle the link. This code is just temporary while we're testing it out internally.

shawaj:
@ChunChiehKung it's as follows:

helium://add_gateway/CrUBCiEBGA3C8JRyKndYev4ShcVFw821Vw7pvZDxiDUHIIiX_50SIQCmsAhGVS5c38NyssphK33KatDkC6h1gqA8gQhjgMJ5uyJIMEYCIQCN4Ouytl7WW92Qt6AHlXSQO-FmfoV3sjoHutEykdCR8wIhAPtVs-qAj75G4IxUMe5f9X5lSqpAvkb6XpAba8RCuMRiKiEBGA3C8JRyKndYev4ShcVFw821Vw7pvZDxiDUHIIiX_504AQ

This is a base 64 hash of the following info I believe:

PR #630, opened 2021-06-02
h3 hexagon support for the network map

### LINKED ISSUE #528: Res 8 Update: Hotspot Details (Phase 1)
Change how we represent Hotspots in the app and coverage map from point location to a resolution 8 hexagon.

This will change how witnesses, followed hotspots, online hotspots will be displayed.

In general, color a hex if it's occupied by a Hotspot.
Purple: followed Hotspot
Yellow: Witness Hotspot
Blue: Owned Hotspot
Navy: Occupied by a Hotspot
Grey: unoccupied

More details found in Res 8 Update in Figma.

tyler-whitman:
closes #528


src/components/Map.tsx
```diff
@@ -299,43 +262,45 @@ const Map = ({
...
-        <MapboxGL.ShapeSource
-          id="ownedHotspots"
-          shape={shapeSources.ownedHotspotFeatures}
-          onPress={onShapeSourcePress}
-        >
-          <MapboxGL.CircleLayer id="markerOwned" style={styles.markerOwned} />
-        </MapboxGL.ShapeSource>
-        <MapboxGL.ShapeSource
+        <H3Grid bounds={mapBounds} visible={showH3Grid} />
+        <HotspotsCoverage
```
matthewcarlreetz, line 266: There seems to be some layering issues here. I wonder if we could somehow simplify by using the `<Coverage/>` component.

src/features/hotspots/root/HotspotSheetHandle.tsx
```diff
@@ -1,55 +1,40 @@
...
-            <Chevron color="#C2C5E4" height={20} width={20} />
-          </TouchableOpacityBox>
-        </>
+      {showHotspotOptions && hotspot && (
+        <FollowButton address={hotspot.address} />
+      )}
+      <Box height={22} justifyContent="center">
+        <CardHandle />
+      </Box>
+      {showHotspotOptions && (
```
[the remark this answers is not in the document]
tyler-whitman: Added a hitSlop

src/components/Map.tsx
tyler-whitman: I think they are all resolved, I would prefer to leave it as is because I like the modular approach of multiple layers so that the coverage component doesn't get inflated with everything from the network map. It will also help with customization down the road is discovery needs changes the network map doesn't. If we find problems I'm open to changing it, but I think its working well.

src/components/H3Grid.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const boundingBox = useMemo(() => {
+    if (zoomLevel < 11) return { type: 'Feature' } as Feature
+    return {
+      type: 'Feature',
+      geometry: {
+        type: 'Polygon',
+        coordinates: bounds
+          ? [
+              [
+                [bounds[0][0], bounds[0][1]],
```
allenan, line 42: this bounding box logic is a little obtuse, and I see it in a couple places. any way to extract it to a utility function?

src/features/hotspots/root/HotspotsView.tsx
```diff
@@ -225,11 +234,28 @@ const HotspotsView = ({
...
   }, [])
 
+  const onMapHexSelected = useCallback(
+    async (hexId: string) => {
+      setSelectedHexId(hexId)
+      setSelectedHotspotIndex(0)
+      const result = (await dispatch(fetchHotspotsForHex({ hexId }))) as {
+        payload?: Hotspot[]
+      }
+      if (result && result.payload && result.payload.length) {
```
allenan, line 244: wondering if you could use the safe accessor pattern for things like this?

`if (result?.payload?.length) {`

src/components/H3Grid.tsx
```diff
@@ -0,0 +1,85 @@
...
+    if (zoomLevel < 11)
+      return { type: 'FeatureCollection', features: [] } as FeatureCollection
+    const hexagons = geojson2h3.featureToH3Set(boundingBox, res)
+    return geojson2h3.h3SetToFeatureCollection(hexagons, (h3Index) => ({
+      id: h3Index,
+    }))
+  }, [boundingBox, res, zoomLevel])
+
+  if (!bounds || !visible) {
+    return null
```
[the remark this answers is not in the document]
tyler-whitman: Yeah I noticed this too, updated it to just return an empty feature set instead

src/components/H3Grid.tsx
tyler-whitman: Extracted it to MapUtils

src/features/hotspots/root/HotspotsView.tsx
tyler-whitman: Updated

matthewcarlreetz (approved):

PR #633, opened 2021-06-02
Feature/625 maker config

### LINKED ISSUE #625: Maker Config
To make it easier for new makers to add their Hotspot to the app.

matthewcarlreetz:
Closes #625


src/makers/hotspot/rak.ts
```diff
@@ -0,0 +1,41 @@
...
+        'Press the black button on your Hotspot. Its light should turn blue.',
+        'Placeholder',
+      ],
+    },
+    ko: {},
+    ja: {},
+    zh: {},
+  },
+  icon: RAK,
+  antenna: { us: RAK_ANTENNAS.RAK_US, default: RAK_ANTENNAS.RAK_EU },
```
tyler-whitman, line 22: Should default be EU?

matthewcarlreetz: I think so. The code checks `const isUS = country === 'US'` and if so it picks the `us` configuration.

tyler-whitman: Hmm maybe we call it EU and the code just defaults to EU? I did it that way because I couldn't find a good way to determine EU since there are many countries. So we just check for US and then if its not use EU since they all had only 2. I guess if there were 3 it would be different.

tyler-whitman: Actually I guess default makes sense haha, I guess it only really becomes an issue if there are more than 2 types which I'm not sure will happen

src/makers/hotspots/index.ts
```diff
@@ -0,0 +1,33 @@
+import longAP from './longAP'
+import rak from './rak'
+import nebra from './nebra'
+import helium from './helium'
+import { LangType, supportedLangs } from '../../utils/i18n/i18nTypes'
+import { HotspotMakerLangField } from './hotspotMakerTypes'
+
+export const HotspotMakerModels = { ...helium, ...rak, ...longAP, ...nebra }
```
tyler-whitman, line 8: We need to add the rest here correct?

tyler-whitman (approved):
Awesome, this is much more simple. So the steps for makers would be add a hotspot file, add an antenna file, and them import it in the index?

Also we just need to add the rest of the current makers right?

src/makers/hotspots/index.ts
matthewcarlreetz: oh, yes, good catch!

PR #634, opened 2021-06-02
Amount locale issues

### LINKED ISSUE #621: Decimal separator conversion issue (e.g., "." to "," and vice versa)
App version 3.2.2 - iOS
Locale = USA / EU

Steps to Reproduce :

- On an app with locale set to US, scan a `payment` QR code with an `amount` using the European decimal separator, `,` (e.g., `312,0919210`).
- The `amount` shown in the payment staging area ignores the decimal separator and instead displays the `amount` as a whole number (e.g., `3,120,919,210`)

- Similarly, on an app with locale set to EU, scan a `payment` QR code with an `amount` using the American decimal separator, `.` (e.g., `312.0919210`).
- The `amount` shown in the payment staging area ignores the decimal separator and instead displays the `amount` as a whole number (e.g., `3,120,919,210`)

Expected Result :

- Respect the decimal separator across regions.

The issue is likely occurring in the `parseAmount` function in `src/utils/transactions.ts`

tyler-whitman:
This fixes two issues with payment amounts and locales

1. On iOS the device region and language had to match for amount numbers to behave properly (EG you can set DE region and English language). Now we use only region to determine amount formatting.
2. There was a bug with scanning QR data if your phone was not in US region which is now resolved. For example this use case: `locale set to EU, scan a payment QR code with an amount using the American decimal separator, . (e.g., 312.0919210). The amount shown in the payment staging area ignores the decimal separator and instead displays the amount as a whole number (e.g., 3,120,919,210)`

closes #621


src/features/wallet/send/SendView.tsx
```diff
@@ -194,11 +193,16 @@ const SendView = ({
       let amount = ''
       let balanceAmount = new Balance(0, CurrencyType.networkToken)
       if (scanAmount) {
+        const floatAmount = parseFloat(scanAmount.split(',').join(''))
```
allenan, line 196: is this just attempting to remove any comma formatting? you could do `scanAmount.replace(/,/g, '')` rather than splitting and rejoining

tyler-whitman: Yeah, sounds good let me update it

tyler-whitman: This is for the QR data, which we will require in US format so we can manually parse out the `,` which parseFloat doesnt work with

allenan (approved):
nice, seems like a better use of the balance class

PR #642, opened 2021-06-04
fix send forum corners

### LINKED ISSUE #608: missing rounded corners when scanning payment_v2 transactions
IMG_752757072907-1

tyler-whitman:
closes #608

### DIFF
--- a/src/components/InputField.tsx
+++ b/src/components/InputField.tsx
@@ -18,6 +18,8 @@ type Props = {
   defaultValue?: string
   value?: string
   numberOfLines?: number
+  bottom?: boolean
+  top?: boolean
 }
 
 const InputField = ({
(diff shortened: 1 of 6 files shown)


src/components/InputField.tsx
```diff
@@ -18,6 +18,8 @@ type Props = {
   defaultValue?: string
   value?: string
   numberOfLines?: number
+  bottom?: boolean
```
matthewcarlreetz, line 21: At a glance what `top` and `bottom` means is unclear to me. Maybe we could call them `topRadius` and `bottomRadius`? Another pattern I've been using for rounding corners is `isFirst` and `isLast`.

matthewcarlreetz (approved):

src/components/InputField.tsx
tyler-whitman: Good point, I like isFirst, isLast. Let me update

PR #643, opened 2021-06-07
Reconfigure file structure for makers to incorporate id and support email

matthewcarlreetz:


src/makers/index.ts
```diff
@@ -0,0 +1,82 @@
...
+  bobcat,
+  finestra,
+  helium,
+  longAP,
+  nebra,
+  rak,
+  syncrobit,
+}
+
+export const HotspotMakerModels = {
```
tyler-whitman, line 22: Couple things here,

1. We need to make sure we keep the original order of the list, which would be ordered by maker ID.
2. I don't see Helium in the hotspot selection list for some reason

tyler-whitman: Maybe its because the syncrobit hotspots.ts file is named Helium?

tyler-whitman: Ah yeah looks like thats why

tyler-whitman: Never mind on the ordering, Coco confirmed alphabetical is ok. Could we make sure it stays sorted no matter the order these are exported in though?

matthewcarlreetz: Great find on the syncrobit!

matthewcarlreetz: The order is currently alphabetical.
```
export const HotspotModelKeys = Object.keys(
  HotspotMakerModels,
).sort() as HotspotType[]
```

tyler-whitman (approved):

PR #650, opened 2021-06-08
Fix wifi showing as connected after forgetting.

### LINKED ISSUE #602: Forget WiFi network asks for password again
When clicking to "forget" a WiFi network in the app, it then takes you to a screen where it asks you to enter the WiFi password for the same network again.

If you click "back" in the app at this point it actually has forgotten the WiFi network and isn't connected.

But if you continue to put the password in it actually reconnects to the same network.

Guess there is a bug somewhere

(Tested using OG and RAK hotspots)

App version 3.2.1
Android 11
Samsung galaxy s21 ultra 5g

matthewcarlreetz:
Closes #602


src/utils/useHotspot.ts
```diff
@@ -218,11 +218,10 @@ const useHotspot = () => {
       encoded,
     )
 
+    dispatch(connectedHotspotSlice.actions.setConnectedHotspotWifi(undefined))
```
matthewcarlreetz, line 221: The response appears to be null which is fine. Always clear the connected wifi if the write succeeds.

tyler-whitman (approved):

PR #659, opened 2021-06-08
Bugfix/disco h3x

### LINKED ISSUE #657: some touch events not registering if it's too close to components

matthewcarlreetz:
closes #657

### DIFF
--- a/src/components/Coverage.tsx
+++ b/src/components/Coverage.tsx
@@ -6,7 +6,7 @@ import MapboxGL, {
   SymbolLayerStyle,
 } from '@react-native-mapbox-gl/maps'
 import { h3ToParent } from 'h3-js'
-import React, { memo, useCallback, useMemo } from 'react'
+import React, { memo, useCallback, useEffect, useMemo } from 'react'
 import { StyleProp } from 'react-native'
 import { Hotspot } from '@helium/http'
 import { Position } from 'geojson'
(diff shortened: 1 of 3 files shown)


src/components/Coverage.tsx
```diff
@@ -56,14 +53,13 @@ const Coverage = ({
...
   onHexSelected,
-  visible = true,
   selectedHexId,
   outlineWidth = 2,
   selectedOutlineWidth = 5,
-  outline = true,
-  showCount = false,
   witnesses,
 }: Props) => {
+  useEffect(() => {}, [witnesses])
```
tyler-whitman, line 61: Why is this needed?

src/components/Coverage.tsx
```diff
@@ -56,14 +53,13 @@ const Coverage = ({
...
+
   const boundingBox = useMemo(() => {
     return boundsToFeature(bounds)
```
tyler-whitman, line 64: This is where I added a check for `if (zoomLevel < 11) return { type: 'Feature' } as Feature` which helps with performance but then you have to set up an effect to keep track of the zoomLevel

tyler-whitman: I also added `    if (!bounds || !visible || zoomLevel < 11)
      return { type: 'FeatureCollection', features: [] } as FeatureCollection` in the source set function

src/components/Coverage.tsx
matthewcarlreetz: forgot to clean it up 🤦

tyler-whitman (approved):

PR #672, opened 2021-06-09
fix assert map crashing

### LINKED ISSUE #671: iOS onboarding issue: screen goes blank after dropping the pin
latest 3.3.0:
on iOS, go through add hotspot flow > get to assert location page > move the pin, wait for it to drop > screen goes blue.

on Android, not reproducible, but unable to complete onboarding flow (bluetooth characteristic error).

tyler-whitman:
The map was starting out zoomed and trying to calculate bounds, we need to prevent calculations when zoomLevel is null

closes #671


danielcolinjames (approved):
looks good to me! 🚢

PR #678, opened 2021-06-10
allow makers to set an URL with wallet

BenoitDuffez:
if the qr translation array contains a second item, it will be transformed to a link with the wallet address

this would implement the proposition I made in #498 where the Helium app would send the user to a maker URL with the wallet address, thus saving the user to copy/paste and hence improving UX

this is optional, if the qr array contains 1 item, nothing is added

example implementation:
```ts
qr: [
  "Generic text to help onboard",
  "https://maker.com/onboard/WALLET|maker.com/onboard"
]
```

Would create a link to `https://maker.com/onboard/14aEqJhfWwvynhHSzP3cLuZuW7yuENkdD3nu3yFZBQmZNpaYMa2`


cokes518:
what happens when the user clicks on the link? it will open in the mobile browser, fill in the wallet address, then the webpage will generate a QR code for them to scan...

but they're still on their phone. How would you solve that?

BenoitDuffez:
The app supports deep links and can receive an add_gateway transaction. So a link on the web page to `helium://add_gateway/txn` would do the trick.

BenoitDuffez:
See: https://github.com/helium/hotspot-app/blob/909e9967d451cf431b25c6fd4a8a08fb6c42b57c/src/providers/AppLinkProvider.tsx#L84-L95

matthewcarlreetz:
@cokes518 @PeterMain Do you want to mock up a design for how this linking will look?

PeterMain:
Sounds like a decent portion is happening 'off-app' - would be good to see some Coco wires, if poss

BenoitDuffez:
@cokes518 congrats on the app release today. Did you consider adding this?
I think it's a huge improvement in UX if the user doesn't have to type in the wallet address on their computer.

BenoitDuffez:
Is there anything I can do for this? The modification seems trivial and harmless to me, let me know if you need me to improve it in any way.

PR #684, opened 2021-06-11
URL-decode the deep link received transaction

BenoitDuffez:
the transaction is sent from an app link, e.g. `helium://add_gateway/XXX`
this path shall not contain any slash, so the incoming transaction (in base64) must be URL-encoded to be safely passed
this change will URL-decode the received transaction so that it is valid base64

e.g.
```
helium://add_gateway/CroBCiEAd9p%2FZOinDi87
```
transaction string is `CroBCiEAd9p%2FZOinDi87`
real base64 string is `CroBCiEAd9p/ZOinDi87`

### DIFF
--- a/src/providers/AppLinkProvider.tsx
+++ b/src/providers/AppLinkProvider.tsx
@@ -92,7 +92,7 @@ const useAppLink = () => {
           const { address: txnStr } = record as AppLink
           if (!txnStr) return
 
-          navigator.confirmAddGateway(txnStr)
+          navigator.confirmAddGateway(decodeURIComponent(txnStr))
           break
         }
       }


timcooijmans:
This is not the right approach; the protobuf should simply be base64url encoded. It was exactly designed for this purpose.

BenoitDuffez:
I know, but the string is decoded as base64. My fix is prob wrong then, it should change the decoder to base64url instead.

BenoitDuffez:
In which case the fix would be to revert my commit and fix here instead: https://github.com/helium/helium-js/blob/master/packages/transactions/src/AddGatewayV1.ts#L86
But changing there would force the change from base64 to base64url everywhere, I think the implications are bigger.

BenoitDuffez:
This seems to be used drop-in for the QR code, so I believe the spec for the QR code would be `helium://add_gateway/txn_in_b64url`
I had initially expected the QR code to be the binary transaction (MUCH shorter than the b64url equivalent, so the QR code is much less complex and thus easier to scan with poorer cameras/lighting conditions).

timcooijmans:
> In which case the fix would be to revert my commit and fix here instead: https://github.com/helium/helium-js/blob/master/packages/transactions/src/AddGatewayV1.ts#L86
> But changing there would force the change from base64 to base64url everywhere, I think the implications are bigger.

There's no need to change from `base64` to `base64url`! Please read the documentation (https://nodejs.org/api/buffer.html#buffer_buffers_and_character_encodings) the `base64` decoder already decodes base64url-encodings.

 

> This seems to be used drop-in for the QR code, so I believe the spec for the QR code would be `helium://add_gateway/txn_in_b64url`
> I had initially expected the QR code to be the binary transaction (MUCH shorter than the b64url equivalent, so the QR code is much less complex and thus easier to scan with poorer cameras/lighting conditions).

The QR-code contains the minimal-size partially signed add-gateway transaction. You seem to suggest that `CroBCiEAd9p/ZOinDi87` (28 bytes) could also be the transaction. I don't know where you get this from, because that's technically impossible. Here you can see the proto: https://github.com/helium/proto/blob/master/src/blockchain_txn_add_gateway_v1.proto. Just the signature of the gateway (ECDSA P-256) is already 64-bytes without any structure-encoding (like DER).

BenoitDuffez:
I didn't know that the decoder did both. Thank you. Problem solved.

My suggestion was just for illustration purposes. It's about 190 bytes actually when complete, a little less for the partial transaction.

PR #685, opened 2021-06-11
hotspot witness list

### LINKED ISSUE #674: unable to see witness list in v3.3.0
With the witness map layer, there's no way for a user to see the Hotspots in its witness list, because every time you tap on a yellow hex, the Hotspot selected changes.

We need to add in a witness list (use the witness list from the chipmunk file in figma)

image

on the witness list, who's the scale, distance (point to point), online/offline, and if it's relayed.

Interaction: tapping on the witness in the list will bring up the Hotspot in the main details, update the map to show the hotspot's hex, and bottom card will go to earnings.

tyler-whitman:
closes #674


src/components/HotspotListItem.tsx
```diff
@@ -68,6 +78,14 @@ const HotspotListItem = ({
     return `${geo.longStreet}, ${geo.longCity}, ${geo.shortCountry}`
   }, [hotspot, t])
 
+  const isRelayed = useMemo(() => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    // TODO: fix listen_addrs in helium-js for witnesses
+    const listenAddrs = hotspot?.status?.listen_addrs
```
tyler-whitman, line 85: Looks like the witness hotspot objects dont have camel case for `listen_addrs`

src/features/hotspots/details/HotspotDetails.tsx
```diff
@@ -317,6 +376,27 @@ const HotspotDetails = ({
...
+            ) : (
+              <>
+                <Box backgroundColor="grayBox" marginBottom="xxs" marginTop="m">
+                  <Text variant="body1Medium" color="grayDarkText" padding="m">
+                    {t('hotspot_details.num_witnesses', {
+                      count: witnesses?.length || 0,
+                    })}
+                  </Text>
+                </Box>
+                {witnesses?.map((witness) => renderWitnessItem(witness))}
```
tyler-whitman, line 395: I had this as a FlatList originally and it works, but since this is already inside a scroll view react throws an error. So I changed it to just directly render them in the scrollview. Open to ideas if you have any.

src/components/HotspotListItem.tsx
matthewcarlreetz: Should we fix helium-js before merging?

src/components/HotspotListItem.tsx
```diff
@@ -130,6 +151,39 @@ const HotspotListItem = ({
                     </Text>
                   </>
                 )}
+                {distanceAway !== undefined && (
+                  <Box marginRight="s" flexDirection="row" alignItems="center">
+                    <LocationIcon color="#474DFF" width={10} height={10} />
```
matthewcarlreetz, line 156: Can we get this from the theme? `<LocationIcon color={colors.purpleMain} width={10} height={10} />`

src/features/hotspots/details/HotspotDetails.tsx
```diff
@@ -28,17 +33,21 @@ import { getSyncStatus, isRelay } from '../../../utils/hotspotUtils'
 import ShareHotspot from '../../../components/ShareHotspot'
 import TouchableOpacityBox from '../../../components/TouchableOpacityBox'
 import Articles from '../../../constants/articles'
+import HotspotListItem from '../../../components/HotspotListItem'
+import { distance } from '../../../utils/location'
```
matthewcarlreetz, line 37: I didn't realize we implemented the haversine formula in the app. I was using a library for disco to do it, but that bit got removed. `"haversine-distance": "^1.2.1"`. We should probably use it or remove it.

src/features/hotspots/details/HotspotDetails.tsx
```diff
@@ -197,6 +211,48 @@ const HotspotDetails = ({
     )
   }, [t])
 
+  const getDistance = useCallback(
+    (otherHotspot: Hotspot) => {
+      if (
```
matthewcarlreetz, line 216: Could use `!hotspotHasValidLocation(hotspot) || !hotspotHasValidLocation(otherHotspots)`

src/features/hotspots/details/HotspotDetails.tsx
```diff
@@ -317,6 +376,27 @@ const HotspotDetails = ({
             />
           </>
         )}
+
+        {selectedOption === 'witnesses' && (
+          <>
+            {hotspotDetailsData.loading ? (
+              <Box marginTop="xl">
+                <ActivityIndicator color="#81909F" />
```
matthewcarlreetz, line 384: `colors.grayMain`

src/features/hotspots/root/HotspotsView.tsx
```diff
@@ -407,6 +407,7 @@ const HotspotsView = ({
           hotspot={selectedHotspot}
           onLayoutHeader={handleDetailHeaderLayout}
           onFailure={handleBack}
+          onSelectHotspot={handlePresentDetails()}
```
matthewcarlreetz, line 410: I'm not sure why this function is curried? Could just be a carryover from a change that was made. Doesn't hurt anything, but could be cleaned up.

src/utils/hotspotUtils.ts
```diff
@@ -37,10 +37,11 @@ export const generateRewardScaleColor = (rewardScale: number): Colors => {
...
 }
 
-export const isRelay = (listenAddrs: string[]) => {
+export const isRelay = (listenAddrs: string[] | undefined) => {
+  if (!listenAddrs) return false
   const IP = /ip4/g
 
-  return !!(
+  return (
     listenAddrs &&
```
matthewcarlreetz, line 45: I think `listenAddrs &&` can safely be removed now that you added the early return.

matthewcarlreetz (approved):

matthewcarlreetz:
IMG_7673
I pretty consistently am getting this error when clicking on witnesses. I think it's because the find in `handlePresentDetails` can return `-1` which breaks `<ContentPill>`. You could just add an early return to the useEffect in `<ContentPill>`

```ts
  useEffect(() => {
    if (selectedIndex < 0) return
    flatListRef?.current?.scrollToIndex({
      animated: true,
      index: selectedIndex,
    })
  }, [selectedIndex])
```

src/features/hotspots/details/HotspotDetails.tsx
matthewcarlreetz: The way you have it feels right to me.

matthewcarlreetz (approved):

src/components/HotspotListItem.tsx
tyler-whitman: I might need to make another change there for explorer, but was planning to merge and fix it later just to get this out since the release is waiting for it.

src/features/hotspots/details/HotspotDetails.tsx
tyler-whitman: Nice helper, but now TS is complaining below that they may be undefined. Know any way to fix that?

src/features/hotspots/root/HotspotsView.tsx
tyler-whitman: Updated it

src/features/hotspots/details/HotspotDetails.tsx
tyler-whitman: Ill just go with out local one, its a pretty simple function

tyler-whitman:
> ```ts
> if (selectedIndex < 0) return
> ```

Added the check and this shouldn't happen now with the fixes from main, but its good to check for it anyway.

PR #687, opened 2021-06-15
Chore/627 integration tests

matthewcarlreetz:
Add tests for number formatting on Send View


src/features/wallet/send/__test__/SendView.german.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+        account: {
+          account: createAccount({
+            balance: new Balance(79942876300, CurrencyType.networkToken),
+          }),
+        },
+      },
+    },
+  )
+  const amountInput = await findByTestId('AmountInput')
+  fireEvent.changeText(amountInput, '1000,35')
```
tyler-whitman, line 70: Nice, this is super cool. Easy to use

src/features/wallet/send/__test__/SendView.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+      },
+    },
+  )
+  const text = await findByText('799.429 HNT Available')
+  expect(text).toBeDefined()
+})
+
+it('comma formats amount separator', async () => {
+  const { findByTestId } = render(
+    <Navigator>
```
tyler-whitman, line 34: Do we need to pass a `Navigator` every time? Maybe we can put that inside the render and make it so that you can just pass the component? Would it support passing props to the component too?

tyler-whitman (approved):
I like the setup. Great start to testing

src/features/wallet/send/__test__/SendView.test.tsx
matthewcarlreetz: In this case, it is a Stack Navigator. My thought was that there might be situations where we want different types of navigators in the future.

PR #694, opened 2021-06-17
lint and ts cleanup

### LINKED ISSUE #695: Fix lint and TS errors
Fix lint and TS errors

tyler-whitman:
requires https://github.com/helium/helium-js/pull/187

- cleans up the lint and ts errors
- adds lint as a build step on github ci

I tried to fix as many as I could but some of the more complicated types I ended up just ignoring for now. If any TS pro wants to take a look that would be great haha. 

closes #695

### DIFF
--- a/src/components/TextTransform.tsx
+++ b/src/components/TextTransform.tsx
@@ -15,7 +15,8 @@ const components = {
 
 type Props = TextProps<Theme> & {
   i18nKey: string
-  values?: Record<string, unknown>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  values?: Record<string, any>
   numberOfLines?: number
   adjustsFontSizeToFit?: boolean
   maxFontSizeMultiplier?: number
--- a/src/components/TextTransform.tsx
+++ b/src/components/TextTransform.tsx
@@ -15,8 +15,7 @@ const components = {
 
 type Props = TextProps<Theme> & {
   i18nKey: string
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  values?: Record<string, any>
+  values?: Record<string, unknown>
   numberOfLines?: number
   adjustsFontSizeToFit?: boolean
   maxFontSizeMultiplier?: number
--- a/src/features/wallet/root/walletLayout.tsx
+++ b/src/features/wallet/root/walletLayout.tsx
@@ -1,10 +1,12 @@
 /* eslint-disable react/jsx-props-no-spreading */
 import { useBottomTabBarHeight } from '@react-navigation/bottom-tabs'
-import React from 'react'
+import React, { ReactElement } from 'react'
 import { SafeAreaInsetsContext } from 'react-native-safe-area-context'
 import { wh, hp } from '../../../utils/layout'
 
-export const withWalletLayout = (WrappedComponent: any) => (props: any) => {
+export const withWalletLayout = (WrappedComponent: ReactElement) => (
+  props: unknown,
+) => {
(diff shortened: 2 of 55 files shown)


danielcolinjames (approved):

src/components/TextTransform.tsx
```diff
@@ -15,7 +15,8 @@ const components = {
 
 type Props = TextProps<Theme> & {
   i18nKey: string
-  values?: Record<string, unknown>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  values?: Record<string, any>
```
matthewcarlreetz, line 19: I think this could be
```ts
  values?: Record<string, never>
```

tyler-whitman: Actually I think unknown would work here, not sure why I changed it. Might have got mixed up in my changes

src/features/wallet/root/walletLayout.tsx
```diff
@@ -4,6 +4,7 @@ import React from 'react'
 import { SafeAreaInsetsContext } from 'react-native-safe-area-context'
 import { wh, hp } from '../../../utils/layout'
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const withWalletLayout = (WrappedComponent: any) => (props: any) => {
```
matthewcarlreetz, line 8: I think we could clean up one of these any's like this
```ts
export const withWalletLayout = (WrappedComponent: React.ComponentType) => (
  props: any,
) => 
```

src/store/hotspotDetails/hotspotDetailsSlice.ts
```diff
@@ -40,16 +40,18 @@ export const fetchHotspotChallengeSums = async (
...
   challengeSums.reverse()
-  const totalSum = challengeSums.reduce(
-    (a, b) => ({ sum: a.sum + b.sum } as any),
-  )
-  const prevSum = challengeSumsPrevious.reduce(
-    (a, b) => ({ sum: a.sum + b.sum } as any),
-  )
+  let totalSum = 0
+  let prevSum = 0
+  challengeSums.forEach((sum) => {
```
matthewcarlreetz, line 45: I think this would be a little more concise way of writing this
```ts
const totalSum = challengeSums.reduce((total, { sum }) => total + sum, 0)
```

src/theme/theme.ts
```diff
@@ -357,6 +357,8 @@ export const darkTheme: Theme = {
     primaryMain: palette.black,
     secondaryMain: palette.blueGray,
   },
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
matthewcarlreetz, line 361: I think we should just remove the dark theme entirely. Doesn't seem to be something we're going to prioritize.

matthewcarlreetz (approved):
Looks good! Just a couple of suggestions.

src/features/wallet/root/walletLayout.tsx
tyler-whitman: It breaks where you use it to wrap a component. I also tried a bunch of others such as `Element`, `JSX.Element`, `ReactElement` with no success :/

src/store/hotspotDetails/hotspotDetailsSlice.ts
tyler-whitman: updated

src/theme/theme.ts
tyler-whitman: Sounds good, didnt seem like it was used anywhere?

PR #709, opened 2021-06-18
Fix support for QR-scanned multi-payments

MichaelBrew:
## Summary
Currently, the QR scanning feature in "Send" supports multiple payments per-transaction if the QR code data is formatted exactly as follows:
```typescript
{
  type: string
  payees: {
    [payeeAddress]: { amount: string, memo?: string },
    [payeeAddress]: { amount: string, memo?: string },
    ...
  }
}
```
However, it does not fully support the following
```typescript
{
  type: string
  payees: {
    [payeeAddress]: amount,
    [payeeAddress]: amount,
    ...
  }
}
```
In the case above, only the address is pre-filled in the "Send" view (amount field remains blank).

This PR adds back support for the latter format which is helpful when submitting a large number of payments (the more payments you can squeeze into a QR code the better, so if you don't need memos then the latter format is more concise). This also updates support for the former format so that amount can be passed as a string or a number (currently, if passed as a number, the app crashes when trying to format it as if it was a string).

### DIFF
--- a/src/providers/AppLinkProvider.tsx
+++ b/src/providers/AppLinkProvider.tsx
@@ -147,7 +147,8 @@ const useAppLink = () => {
    * (1) A helium deeplink URL
    * (2) address string
    * (3) stringified JSON object { type, address, amount?, memo? }
-   * (4) stringified JSON object { type, payees: {[payeeAddress]: { amount, memo? }} }
+   * (4) stringified JSON object { type, payees: {[payeeAddress]: amount} }
+   * (5) stringified JSON object { type, payees: {[payeeAddress]: { amount, memo? }} }
    */
   const parseBarCodeData = useCallback(
     (data: string, scanType: AppLinkCategoryType): AppLink | AppLinkPayment => {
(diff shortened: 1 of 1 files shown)


src/providers/AppLinkProvider.tsx
```diff
@@ -211,11 +212,26 @@ const useAppLink = () => {
...
+                    }
+                    amount = scanData.amount
+                    memo = scanData.memo
+                  }
+                }
                 return {
                   address: entries[0],
-                  amount: scanData.amount,
-                  memo: scanData.memo,
+                  amount: `${amount}`,
```
MichaelBrew, line 233: If `amount` was actually defined as a number in the QR code JSON data, then this later triggers a crash due to a JS runtime error (in production you see a blank, dark-blue screen, in development you see the screen below)
image

tyler-whitman: Nice good catch, thanks for fixing this

rawrmaan:
Really looking forward to this. Keep needing to revert to older versions on TestFlight to do our host payouts :)

tyler-whitman:
> Really looking forward to this. Keep needing to revert to older versions on TestFlight to do our host payouts :)

I think we are safe to get this in to the next release. If you are on the Testflight version of the app could you give it a test for your use case?

rawrmaan:
> > Really looking forward to this. Keep needing to revert to older versions on TestFlight to do our host payouts :)
> 
> I think we are safe to get this in to the next release. If you are on the Testflight version of the app could you give it a test for your use case?

Tested, it works! Thanks Tyler.

PR #710, opened 2021-06-18
ensure all invalid quote occurrences get replaced

tyler-whitman:
replace would only get the first one, which actually fixed the bug but we should replace all of them


matthewcarlreetz (approved):
What was the issue here?

cokes518:
I used a slanty quotation mark instead of straight in a notification and it crashed the app.
________________________________

Issue #133, opened 2021-02-01
Assert Hotspot Improvements: Scan QR code with location


rawrmaan:
Add the ability to scan a QR code when updating a location. The QR encoded data would be JSON encoded and contain either an `h3_location` or `lat` and `lng` coordinates which could be converted to H3 within the app. This would bring up the assert location UI with its map centered on that location.

Future improvements: Once HIP 12 and `assert_location_v2` are implemented, add a hotspot `address` field to the JSON object to allow one-tap scanning and updating of locations without having to navigate to the hotspot's page first.

PR #730, opened 2021-06-27
[#133] Scan QR Code to Reassert Hotspot Location

MichaelBrew:
Feature request: https://github.com/helium/hotspot-app/issues/133

## Summary
This PR adds the ability to scan a QR code (structure defined below) to reassert the location of a hotspot within a single screen. The code can be scanned by tapping the QR symbol within the Wallet View, and the resulting screen will include: (1) the selected hotspot's name, (2) the cost to reassert, (3) a preview of the new location on a map, (4) the name of the location (city/street/country), (5) and CTAs to cancel or confirm.

**QR Code JSON Payload**
```json
{
  "address": "11vAu1u...",
  "lat": 32.783364464812564,
  "lng": -96.78312503986781,
}
``` 

| Screenshot | Invalid Address | Demo* |
| ---- | ---- | ---- |
| image | Snip20210821_3 | hotspot_location_qr_code_demo |

_* To facilitate the demo on local simulator, I hooked up the "X" icon from the Scan View into `handleBarCodeScanned` with a hardcoded JSON payload. Also, I haven't been able to get MapBox to work locally so no map is shown on the confirmation screen._

## Implementation
This PR largely relies on a new `HotspotLocationUpdateScreen` component nested within `HotspotsStack` stack navigator (alongside `HotspotsScreen`). The hotspot address and location parsed from the QR code are passed to this screen via navigation props. It then renders an updated version of `ReassertLocationFee` which accepts three new props:
```javascript
// Parents can optionally pass a new location directly to ReassertLocationFee which will show
// the new location and render in a "Confirmation" state, rather than showing the hotspot's
// current location and the ability to find a new location.
newLocation?: { latitude: number; longitude: number; name?: string }

// If rendering the "Confirmation" state mentioned above, a secondary "cancel" CTA is displayed
// alongside the "I Confirm" primary CTA, triggering this callback
onCancel?: () => void

// Parents can pass "isLoading" to indicate whether they're processing an async action, which
// will be manifested by a loading indicator inside the primary confirmation CTA
isLoading?: boolean
```
This allows `ReassertLocationFee` to act as both a one-stop "Confirmation" screen for reasserting locations, as well as an entry point into the usual reassert flow (show fee and current location, and allow users to select a new location).

The "assert location" transaction logic is largely copied from `UpdateHotpostConfig`. Given the transaction-specific logic is abstracted within `assertLocationTxn` and `submitTxn`, the remaining copypasta is mainly success/error handling which didn't feel necessary or appropriate to extract into a common location and share across different screens (mainly uses Toast, Alert, and Logger).

## Testing
So far the extent of testing done is basically what's shown in the demo above. The `submitTxn` call has been stubbed out (for an empty Promise that takes ~1.5sec to resolve), relying on the assumption that `assertLocationTxn` and `submitTxn` already play nicely (given they're used in the same way during the regular location update flow).

Have also tested that providing either an invalid address in the QR code payload or a hotspot address that doesn't belong to the currently authenticated user will result in the "error" dialog shown above.


timcooijmans:
Please add optional support for height and antenna gain.

tyler-whitman (approved):
Looks great!

PR #731, opened 2021-06-28
Feature/qr onboard link

matthewcarlreetz:
@BenoitDuffez Here are my proposed changes.


allenan (approved):

BenoitDuffez:
The link text is the full URL right? Would it be more elegant to remove http:// and the wallet address? Like `https://maker.com/onboard?wallet=XXX` becomes `maker.com/onboard`

PR #732, opened 2021-06-28
Handle onboarding service unavailable when onboarding hotspot.

### LINKED ISSUE #722: Onboarding Server Error check
The onboarding server went down today, but folks were shown error messages that their Hotspot doesn't exist on the Onboarding server (incorrect).
image
image

We should have a check to see if the onboarding server is running, and if it's not, display a different error message:

Onboarding Error

The Onboarding Server is temporarily unavailable and users won't be able to add Hotspots at this time. 
Check for updates on status.helium.com and try again later.

matthewcarlreetz:
closes #722

### DIFF
--- a/src/features/hotspots/setup/OnboardingErrorScreen.tsx
+++ b/src/features/hotspots/setup/OnboardingErrorScreen.tsx
@@ -29,7 +29,7 @@ const OnboardingErrorScreen = () => {
 
         <Text variant="h1">{t('hotspot_setup.onboarding_error.title')}</Text>
         <Text variant="body1" marginVertical="l">
-          {t('hotspot_setup.onboarding_error.subtitle')}
+          {t(`hotspot_setup.onboarding_error.subtitle.${connectStatus}`)}
         </Text>
         <Text variant="body2" marginVertical="l">
           {connectStatus}
--- a/src/features/hotspots/setup/OnboardingErrorScreen.tsx
+++ b/src/features/hotspots/setup/OnboardingErrorScreen.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react'
+import React, { useCallback, useMemo } from 'react'
 import { RouteProp, useNavigation, useRoute } from '@react-navigation/native'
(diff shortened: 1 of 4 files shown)


src/features/hotspots/setup/OnboardingErrorScreen.tsx
```diff
@@ -29,7 +29,7 @@ const OnboardingErrorScreen = () => {
 
         <Text variant="h1">{t('hotspot_setup.onboarding_error.title')}</Text>
         <Text variant="body1" marginVertical="l">
-          {t('hotspot_setup.onboarding_error.subtitle')}
+          {t(`hotspot_setup.onboarding_error.subtitle.${connectStatus}`)}
```
allenan, line 32: is there a fallback message if there's no i18n string defined for the `HotspotConnectStatus`?

src/utils/useHotspot.ts
```diff
@@ -186,8 +187,16 @@ const useHotspot = () => {
     await updateHotspotStatus(payload?.hotspot)
 
     if (!payload?.onboardingRecord?.onboardingKey) {
-      Logger.error(new Error('Hotspot connect failed: no onboardingKey'))
-      return handleConnectStatus('no_onboarding_key')
+      let err: HotspotConnectStatus = 'service_unavailable'
+      if (payload?.onboardingRecord?.code === 404) {
```
allenan, line 191: does `no_onboarding_key` still occur? I see previously this returned `handleConnectStatus('no_onboarding_key')`, but now I see it returning `invalid_onboarding_address`. Additionally the two subtitle options in the i18n file are `no_onboarding_key` and `service_unavailable` but I see `invalid_onboarding_address` and `service_unavailable` being returned from here

src/features/hotspots/setup/OnboardingErrorScreen.tsx
matthewcarlreetz: There isn't, but there should be.

allenan (approved):

PR #733, opened 2021-06-28
Fix incorrect byte size calc failing memo validation

### LINKED ISSUE #725: Unable to burn HNT
Unable to scan QR code from console to burn HNT. The burn button is greyed out and not selectable.

Reproducible by community on 3.3.1 and me (on 3.3.2)

IMG_CCC068EC78A3-1

matthewcarlreetz:
Closes #725

### DIFF
--- a/src/utils/transactions.ts
+++ b/src/utils/transactions.ts
@@ -198,11 +198,9 @@ export const stringAmountToBalance = (formAmount: string) => {
   return Balance.fromFloat(floatAmount, CurrencyType.networkToken)
 }
 
-export const getMemoBytesLeft = (
-  memo: string,
-): { numBytes: number; valid: boolean } => {
-  if (!memo) return { numBytes: 8, valid: true }
-  const buff = Buffer.from(memo)
+export const getMemoBytesLeft = (base64Memo?: string) => {
+  if (!base64Memo) return { numBytes: 8, valid: true }
+  const buff = Buffer.from(base64Memo, 'base64')
   const size = buff.byteLength
   return { numBytes: size < 8 ? 8 - size : 0, valid: size <= 8 }
 }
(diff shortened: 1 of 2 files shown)


tyler-whitman (approved):

src/utils/transactions.ts
```diff
@@ -198,11 +198,13 @@ export const stringAmountToBalance = (formAmount: string) => {
   return Balance.fromFloat(floatAmount, CurrencyType.networkToken)
 }
 
-export const getMemoBytesLeft = (
-  memo: string,
-): { numBytes: number; valid: boolean } => {
+export const getMemoBytesLeft = (memo: string, shouldEncode = false) => {
```
allenan, line 201: generally I try to prefer not adding options as top level arguments to a function because when the function is used, it's not clear what the argument controls. For example, you'd need to look at the definition of the function to see what the `true` in `getMemoBytesLeft(memo, true)` is telling the function to do. You could maybe use an options argument instead which would be self-documenting: `getMemoBytesLeft(memo, { shouldEncode: true })`. I've also been trying to prefer React components over utility functions in the explorer because props are the equivalent of an options object although that may not be the right solution here

src/utils/transactions.ts
```diff
@@ -198,11 +198,13 @@ export const stringAmountToBalance = (formAmount: string) => {
...
   if (!memo) return { numBytes: 8, valid: true }
-  const buff = Buffer.from(memo)
+  let testStr = memo
```
allenan, line 203: is `testStr` the correct variable name for this?

src/utils/transactions.ts
```diff
@@ -198,11 +198,13 @@ export const stringAmountToBalance = (formAmount: string) => {
...
+  if (shouldEncode) {
+    testStr = encodeMemoString(testStr) || ''
+  }
+  const buff = Buffer.from(testStr, 'base64')
```
allenan, line 207: will this always be constructed from a base64 string? or only in the `shouldEncode` logic branch?

src/utils/transactions.ts
matthewcarlreetz: Totally agree that options are a better way to go. I ended up removing the encoding from this utility function altogether. I think it's much clearer now.

PR #736, opened 2021-06-28
Add conditions to reduce status banner display frequency

### LINKED ISSUE #735: Improve outages experience
in the app, whenever there is an active incident on status page, we display a banner at the top every time a user opens the app, regardless of whether they dismissed it.

1. we should remember if a user has dismissed an incident, and do not display the banner again (for the same incident).

2. we should also make the touch target larger when dismissing the banner

3. lastly, if we can only show specific components from status, we can customize app-specific incidents that is relevant to the app user. 

I.e. a console outage is meaningless to an app user, but it will affect discovery mode which users should know about.

matthewcarlreetz:
Remove ui concerns from slice to status component.
Only show relevant incidents.
Persist incidents. Only show ones we haven't before.
Remove status fetching and data.
Localize banner description.

Closes #735

### DIFF
--- a/src/store/helium/heliumStatusSlice.ts
+++ b/src/store/helium/heliumStatusSlice.ts
@@ -2,36 +2,39 @@ import { createAsyncThunk, createSlice } from '@reduxjs/toolkit'
 import AsyncStorage from '@react-native-community/async-storage'
 import { getIncidents } from '../../utils/StatusClient'
 import { Incident, StatusInfo } from './StatusTypes'
+import { getWallet } from '../../utils/walletClient'
 
 const STATUS_KEY = 'statusKey'
-const ONBOARDING_COMPONENT_ID = '74lkk1qwp5xq'
-const APP_COMPONENT_ID = 'qfcq2xt6v0xm'
-const COMPONENT_IDS = [ONBOARDING_COMPONENT_ID, APP_COMPONENT_ID]
-const INCIDENT_STATUSES = [
-  'investigating',
-  'identified',
-  'scheduled',
-  'in_progress',
-  'verifying',
-]
-
 type StoredIncident = {
   id: string
   status: string
 }
 
+type StatusConfig = { components: string[]; incidentStatuses: string[] }
 export type HeliumStatusState = {
   page?: StatusInfo
   incidents: Incident[]
+  config?: StatusConfig
 }
 
 const initialState: HeliumStatusState = {
   incidents: [],
 }
 
+export const fetchStatusConfig = createAsyncThunk<StatusConfig>(
+  'status/getConfig',
+  async () => getWallet('status', null, { camelCase: true }),
+)
+
 export const fetchIncidents = createAsyncThunk(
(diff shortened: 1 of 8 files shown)


src/store/helium/heliumStatusSlice.ts
```diff
@@ -1,67 +1,85 @@
...
-    incidents?: Incident[]
-  }
-}
-const initialState: HeliumStatusState = {
-  status: {},
-  incidents: {},
+const STATUS_KEY = 'statusKey'
+const ONBOARDING_COMPONENT_ID = '74lkk1qwp5xq'
+const APP_COMPONENT_ID = 'qfcq2xt6v0xm'
+const COMPONENT_IDS = [ONBOARDING_COMPONENT_ID, APP_COMPONENT_ID]
```
allenan, line 9: will we want this to be configurable via wallet api? or are we ok with it being hardcoded and releasing a new app version to update these in the future?

src/store/helium/heliumStatusSlice.ts
```diff
@@ -1,67 +1,85 @@
...
-      state.incidents = payload
-      const lastIncident = payload?.incidents[0]
-      if (lastIncident?.impact && lastIncident?.impact !== 'none') {
-        showMessage({
-          type: lastIncident.impact === 'critical' ? 'danger' : 'warning',
-          message: lastIncident.name,
-          description: getAlertDescription(lastIncident.updated_at),
-        })
-      }
+      const incidents = payload.incidents.flatMap((incident) => {
```
allenan, line 57: I haven't seen `flatMap` used this way before. it certainly works. you could also use `reduce` and conditionally add to the accumulator

src/store/helium/heliumStatusSlice.ts
```diff
@@ -1,67 +1,85 @@
...
+        ) {
+          return []
+        }
+        return [incident]
+      })
+
+      const nextStoredIncidents = [
+        ...incidents.map(({ id, status }) => ({ id, status })),
+        ...payload.storedIncidents,
+      ].slice(0, 20) // limit number of stored incidents. This is just an arbitrary number. May need adjustment.
```
allenan, line 77: good that this is limited 👍

allenan (approved):
looks good to me, just a few questions ^

src/store/helium/heliumStatusSlice.ts
matthewcarlreetz: This was actually a trick I found in the Elixir hex docs a little while back. I kinda like it 🤷

src/store/helium/heliumStatusSlice.ts
matthewcarlreetz: Good question. I don't imagine this will change often but will check with Coco.

src/store/helium/heliumStatusSlice.ts
allenan: fair enough, it works :)

allenan (approved):

PR #746, opened 2021-06-30
3.4.0

matthewcarlreetz:
Looking to merge the current state of 3.4.0 into main.


src/features/hotspots/root/HotspotsList.tsx
```diff
@@ -1,21 +1,35 @@
 import React, { memo, useCallback, useMemo } from 'react'
-import { BottomSheetSectionList } from '@gorhom/bottom-sheet'
+import { SectionList } from 'react-native'
```
tyler-whitman, line 2: Does this need to be imported from `react-native-gesture-handler`?

tyler-whitman: Ah its not in a bottom sheet anymore right? So probably not then.

src/features/hotspots/root/HotspotsList.tsx
```diff
@@ -113,15 +127,31 @@ const HotspotsList = ({
...
-    <BottomSheetSectionList
-      sections={sections}
-      keyExtractor={(item: Hotspot) => item.address}
-      ListHeaderComponent={<WelcomeOverview />}
-      renderSectionHeader={renderHeader}
-      renderItem={renderItem}
-      contentContainerStyle={contentContainerStyle}
-      showsVerticalScrollIndicator={false}
-    />
+    <Box backgroundColor="white" flex={1} top={visible ? 0 : wh}>
```
tyler-whitman, line 130: The iOS native status bar time and service text is in dark mode (so that its white) on this screen, but when you have the list open you can no longer see this text. I think it should be in normal mode (black text) instead?

src/features/hotspots/root/HotspotsViewHeader.tsx
```diff
@@ -46,7 +46,7 @@ const HotspotsViewHeader = ({
   const style = useAnimatedStyle(
     () => ({
       position: 'absolute',
-      bottom: showDetails ? -100 : -240,
+      bottom: showDetails ? -100 : -320,
```
tyler-whitman, line 49: This sits a little higher above the card now, not sure if thats intended?

src/features/hotspots/details/HotspotDetails.tsx
```diff
@@ -265,190 +295,239 @@ const HotspotDetails = ({
...
-              numberOfLines={1}
-              adjustsFontSizeToFit
-            >
-              {formattedHotspotName[0]}
-            </Text>
-            <Box flexDirection="row" alignItems="flex-start">
+    <BottomSheet
+      snapPoints={snapPoints}
+      ref={listRef}
+      index={0}
```
tyler-whitman, line 340: One thing that might be nice is to retain the snap index. So if for example you select a hotspot from your list and expand the details, when you hit the home icon to go back to the list and select another the card is in the last state you had it (eg it remains expanded or collapsed)

Not sure if that makes sense or if its intended to reset everytime?

src/features/hotspots/root/ShortcutNav.tsx
```diff
@@ -0,0 +1,394 @@
...
+      onItemSelected(nextItem)
+    },
+    [data, onItemSelected, scrollOffsets, selectedItem],
+  )
+  return (
+    <Box height={SHORTCUT_NAV_HEIGHT} backgroundColor="primaryBackground">
+      <Box top={-43} left={0} right={0} height={43} position="absolute">
+        <Mask width="100%" color={colors.primaryBackground} />
+      </Box>
+      <FlatList
```
tyler-whitman, line 375: This works fantastically, good math for all those calculations haha

src/features/moreTab/more/MoreScreen.tsx
```diff
@@ -295,7 +295,7 @@ const MoreScreen = () => {
   )
 
   return (
-    <SafeAreaBox backgroundColor="primaryBackground">
+    <SafeAreaBox backgroundColor="primaryBackground" flex={1}>
```
tyler-whitman, line 298: I noticed this screen sometimes doesn't display the entire list and gets cut off half way, maybe its related to this change? Not sure if this is a new or existing bug or why you had to change the flex here.

tyler-whitman (approved):
Awesome, working really well. Left some comments but nothing major.

src/features/hotspots/root/HotspotsList.tsx
matthewcarlreetz: Yeah, right, it's just a regular View now.

src/features/hotspots/root/ShortcutNav.tsx
matthewcarlreetz: ha, thanks! It was a bit tricky to figure out.

src/features/hotspots/root/HotspotsViewHeader.tsx
matthewcarlreetz: Good call. I shrunk the ShortcutNav height and forgot to adjust this.

src/features/hotspots/details/HotspotDetails.tsx
matthewcarlreetz: I hear what you're saying. Let's see what others think before making the change.

src/features/moreTab/more/MoreScreen.tsx
matthewcarlreetz: I changed the tabs so they load lazily. The flex was a fix needed because of that. There must still be a little bug here though. Will fix.

src/features/hotspots/root/HotspotsList.tsx
matthewcarlreetz: I noticed this too but forgot to make a change. I'd like it to adjust per tab. I'll figure it out...

src/features/moreTab/more/MoreScreen.tsx
matthewcarlreetz: Setting `initialNumToRender={100}` seems to fix the bug. I also cleaned up the ts warning here.

PR #761, opened 2021-07-07
update notification list font size

tyler-whitman:
- increase list item font size
- move notification details back to bottom

### DIFF
--- a/src/features/notifications/NotificationShow.tsx
+++ b/src/features/notifications/NotificationShow.tsx
@@ -51,7 +51,7 @@ const NotificationShow = ({ notification, onClose }: Props) => {
   )
 
   const bodyStyle = useMemo(
-    () => ({ maxHeight: Platform.OS === 'ios' ? 450 : 350 }),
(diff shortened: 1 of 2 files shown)


src/features/notifications/NotificationShow.tsx
```diff
@@ -51,7 +51,7 @@ const NotificationShow = ({ notification, onClose }: Props) => {
   )
 
   const bodyStyle = useMemo(
-    () => ({ maxHeight: Platform.OS === 'ios' ? 420 : 320 }),
+    () => ({ maxHeight: Platform.OS === 'ios' ? 450 : 350 }),
```
matthewcarlreetz, line 54: This is totally fine as is, but I recently realized there's also a `Platform.select` function that I hadn't been utilizing.

So this could be...
```ts
  const bodyStyle = useMemo(
    () => ({ maxHeight: Platform.select({ ios: 450, android: 350 }) }),
    [],
  )
```

matthewcarlreetz (approved):

PR #762, opened 2021-07-07
check for no location change when updating hotspot location

### LINKED ISSUE #744: Check if location changed in assert_v2 transactions
When a user neglects to update the hotspot location (or forgets, or didn't realize nothing changed), we should let them know and not allow them to submit the transaction.

Because nothing has changed, when the assert_v2 transaction submits, the staking fee is set at 0. This fee is incorrect (it expects 55,000 DC for gain/elevation change or 1,000,000 DC for location change) so the blockchain rejects the transaction and a failed notification is sent.

In the UI, if the user tries to submit an assert_v2 location change with the same lat/long, display an alert:

"Location Unchanged

The Hotspot location has not changed. Drag the pin to a different location and try again.

OK"

tyler-whitman:
closes #744

### DIFF
--- a/src/features/hotspots/settings/ReassertLocationUpdate.tsx
+++ b/src/features/hotspots/settings/ReassertLocationUpdate.tsx
@@ -2,7 +2,8 @@ import React, { memo, useCallback, useEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { Position } from 'geojson'
 import Search from '@assets/images/search.svg'
-import { Platform } from 'react-native'
+import { Alert, Platform } from 'react-native'
+import { Hotspot, Witness } from '@helium/http'
 import Box from '../../../components/Box'
 import Text from '../../../components/Text'
 import Map from '../../../components/Map'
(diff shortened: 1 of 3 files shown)


src/features/hotspots/settings/ReassertLocationUpdate.tsx
```diff
@@ -53,6 +57,14 @@ const ReassertLocationUpdate = ({
 
   const handleNextButton = () => {
     if (!confirming) {
+      const h3Location = getH3Location(markerCenter[1], markerCenter[0])
+      if (h3Location === hotspot?.location) {
+        Alert.alert(
```
matthewcarlreetz, line 62: You could use our `useAlert` hook here.

```ts
        showOKAlert({
          titleKey:
            'hotspot_setup.add_hotspot.assert_loc_error_no_change_title',
          messageKey:
            'hotspot_setup.add_hotspot.assert_loc_error_no_change_body',
        })
```

matthewcarlreetz (approved):

PR #776, opened 2021-07-13
Allow self as discovery responder.

### LINKED ISSUE #723: Disco Mode: Let users know that the only responder is themselves
We made a decision earlier on to show "unable to initiate session" even if the only responder is the Hotspot itself. This creating some confusion so we should let them know that everything worked, but no one else heard them.

So instead of showing the error, we should show them the result with the Hotspot as a responder.

matthewcarlreetz:
Closes #723


src/features/hotspots/settings/discovery/dedupeDiscoveryResponses.ts
```diff
@@ -0,0 +1,9 @@
+import { DiscoveryResponse } from '../../../../store/discovery/discoveryTypes'
+
+export default (responses: DiscoveryResponse[]) =>
+  responses.filter(
```
matthewcarlreetz, line 4: This is the bulk of the change. We used to filter out "self" here.

tyler-whitman (approved):

src/features/hotspots/settings/discovery/dedupeDiscoveryResponses.ts
allenan: would [lodash/uniqBy](https://lodash.com/docs/4.17.15#uniqBy) be useful here? sometimes they use clever algorithms under the hook for these utilities -- was just thinking that filter + findIndex might create a few loops through the set.

matthewcarlreetz: Yeah, I think so. If nothing else, it makes the code more readable.

src/features/hotspots/settings/discovery/dedupeDiscoveryResponses.ts
```diff
@@ -0,0 +1,5 @@
+import { uniqBy } from 'lodash'
+import { DiscoveryResponse } from '../../../../store/discovery/discoveryTypes'
+
+export default (responses: DiscoveryResponse[]) =>
+  uniqBy(responses, (response) => response.hotspotAddress)
```
allenan, line 5: i think you can also do this like `uniqBy(responses, 'hotspotAddress')`

matthewcarlreetz: oh, cool. I've never used it like that before.

allenan (approved):
looks good. one little lodash syntax suggestion above ^

src/features/hotspots/settings/discovery/dedupeDiscoveryResponses.ts
matthewcarlreetz: For TS I think I prefer `uniqBy(responses, ({ hotspotAddress }) => hotspotAddress)`
That way we don't lose type safety.

PR #777, opened 2021-07-13
Change block sync buffer to 1,500.

matthewcarlreetz:
Change block sync buffer to 1,500.
Use hotspot.status.timestamp to determine currently synced block.

### DIFF
--- a/src/features/hotspots/root/HotspotsList.tsx
+++ b/src/features/hotspots/root/HotspotsList.tsx
@@ -1,4 +1,4 @@
-import React, { memo, useCallback, useMemo } from 'react'
+import React, { memo, useCallback, useEffect, useMemo } from 'react'
 import { SectionList } from 'react-native'
 import { Hotspot, Witness } from '@helium/http'
 import { useSelector } from 'react-redux'
@@ -13,11 +13,15 @@ import HotspotListItem from '../../../components/HotspotListItem'
 import { RootState } from '../../../store/rootReducer'
 import WelcomeOverview from './WelcomeOverview'
 import HotspotsPicker from './HotspotsPicker'
-import { HotspotSort } from '../../../store/hotspots/hotspotsSlice'
+import {
+  fetchSyncStatus,
+  HotspotSort,
+} from '../../../store/hotspots/hotspotsSlice'
 import TouchableOpacityBox from '../../../components/TouchableOpacityBox'
 import { wh } from '../../../utils/layout'
 import FocusAwareStatusBar from '../../../components/FocusAwareStatusBar'
 import HotspotsEmpty from './HotspotsEmpty'
+import { useAppDispatch } from '../../../store/store'
 
 const HotspotsList = ({
   onSelectHotspot,
@@ -35,6 +39,7 @@ const HotspotsList = ({
   onRequestShowMap?: () => void
 }) => {
   const colors = useColors()
+  const dispatch = useAppDispatch()
   const { top } = useSafeAreaInsets()
   const loadingRewards = useSelector(
     (state: RootState) => state.hotspots.loadingRewards,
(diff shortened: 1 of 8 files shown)


cokes518:
use a hotspot's  "status": {
      "timestamp": "2021-07-13T04:31:13.478000Z",

to look up gossiped block height:
https://api.helium.io/v1/blocks/height/?max_time=2021-07-13T04:31:13.478000Z

if the height is within 1500 to head, it is synced.

a more elegant solution is TBD, will be part of 3.4.1

tyler-whitman (approved):

src/store/hotspots/hotspotsSlice.ts
```diff
@@ -275,6 +283,53 @@ export const unfollowHotspot = createAsyncThunk<Hotspot[], string>(
...
+      hotspots: {
+        syncStatuses: Record<string, CacheRecord<HotspotStatus>>
+      }
+    }
+
+    if (syncStatuses[address] && hasValidCache(syncStatuses[address])) {
+      throw new Error(`sync status for hotspot ${address} is already valid`)
+    }
+
+    if (!statusTime) {
```
allenan, line 319: status timestamp will be null until it has been gossiped. do we want this to throw an error in that case? or just consider it syncing?

cokes518: I'd consider it syncing

src/features/hotspots/root/HotspotsList.tsx
```diff
@@ -46,9 +51,27 @@ const HotspotsList = ({
...
+    (state: RootState) => state.heliumData.blockHeight,
+  )
+  const syncStatuses = useSelector(
+    (state: RootState) => state.hotspots.syncStatuses,
+  )
 
   const { t } = useTranslation()
 
+  useEffect(() => {
+    orderedHotspots.forEach(({ address, status }) => {
```
allenan, line 64: this seems like it could do a lot of fetching. one thing I talked about with @madninja was that this solution won't be great for lists of hotspots, only for a single hotspot details. because it introduces an n+1 query situation. I see that we show sync status for your hotspot list, and I'm just a little worried about wallets with 100, 300 etc hotspots in them. I know there's a better solution coming that would just deal with block heights so we wouldn't have to do the intermediary timestamp --> block height lookup. Do you think we should wait for that in the app and just go with the 1500 change for now? or is the fetch performance not too bad?

tyler-whitman: Yeah I was thinking the same thing and chatted with Matt, since its temporary I figured it's ok for now but we may want to handle failed states so that it isn't just stuck loading or showing an incorrect sync state when things don't load properly.

tyler-whitman: We could even maybe fallback to the gossiped height?

matthewcarlreetz: I found a hotspot `strong-lipstick-rooster` that has 40 witnesses, the sync statuses load quickly. I chose to not show a loading indicator while they are being fetched. Status will just be blank until it is known.

I could see not loading the sync status of witnesses though. @cokes518 thoughts?

src/store/hotspots/hotspotsSlice.ts
matthewcarlreetz: So, like this?
```ts
    if (!statusTime) {
      return { status: SyncStatus.none, percent: 0 }
    }
```

src/features/hotspots/root/HotspotsList.tsx
matthewcarlreetz: I think falling back to the gossiped height might cause even more confusion 🤷

tyler-whitman: That’s true haha. Maybe it’s fine for now. I think maybe not showing witness status might be a good idea to save all the api calls though. Some have hundreds of witnesses

cokes518: fine with me.

matthewcarlreetz: ok. I'll remove witness status fetching...

PR #779, opened 2021-07-13
Fix all maker support emails going to helium.

matthewcarlreetz:


src/makers/index.ts
```diff
@@ -81,10 +81,6 @@ export const AntennaModelKeys = Object.keys(
 export const AntennaTypeCount = AntennaModelKeys.length
 
 export const getMakerSupportEmail = (makerId?: number): string => {
-  Object.keys(Makers).forEach((m) => {
-    const maker = Makers[m]
-    if (maker.id === makerId) return maker.supportEmail
```
matthewcarlreetz, line 86: This was the root of the issue. `return` in a `forEach` loop is the equivalent of `continue`.

tyler-whitman (approved):

src/makers/index.ts
tyler-whitman: Thanks for fixing this one, I forgot I was going to take a look at it.

PR #788, opened 2021-07-15
Fix the damn lag

### LINKED ISSUE #784: Android 3.4.0 App Performance
User with 68 Hotspots:
LG V60 android
android 11. kernal 4.19.113

Says side swiping with shortcut nav is bad, app unusable.

User with 20 Hotspots
Samsung S20
He's sending me a video of performance.

matthewcarlreetz:
Closes #784 

Render all ShortcutNav pills at once to avoid snap jank.
Only dispatch search if term has changed.
Remove fetch and display of sync status for hotspot list items.
Cleanup HotspotDetailChart layout.
Only load data when bottom sheet is expanded.
Fix offset for last item in shortcutnav.

### DIFF
--- a/src/components/HotspotListItem.tsx
+++ b/src/components/HotspotListItem.tsx
@@ -10,7 +10,7 @@ import { Pressable } from 'react-native'
 import Box from './Box'
 import Text from './Text'
 import useCurrency from '../utils/useCurrency'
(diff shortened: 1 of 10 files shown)


allenan (approved):
code looks good on this. we can have mods test performance in the release build on TF

PR #805, opened 2021-07-16
Fix next button incorrectly disabled when onboarding from qr/web.

matthewcarlreetz:


src/features/hotspots/setup/HotspotSetupExternalConfirmScreen.tsx
```diff
@@ -189,7 +189,7 @@ const HotspotSetupExternalConfirmScreen = () => {
         mode="contained"
         variant="primary"
         onPress={navNext}
-        disabled={publicKey !== address}
+        disabled={ownerAddress !== address}
```
tyler-whitman, line 192: So this was already an existing bug right? Just should have been owner instead of hotspot address?

tyler-whitman (approved):

PR #809, opened 2021-07-19
Feature/801 fleet mode

### LINKED ISSUE #801: Create "Fleet Mode" for App users with many many Hotspots
Reported by Android user on 3.4.1 with 100+ Hotspots in the account, using Samsung Galaxy Tab.

https://explorer.helium.com/accounts/139DUUzVJcR6NcvjkXGJzSpaWLXUjRDvGMEQuHCDkikybzePfn2

matthewcarlreetz:
Closes #801


src/features/hotspots/root/HotspotsScreen.tsx
```diff
@@ -46,12 +87,8 @@ const HotspotsScreen = () => {
   })
 
   useEffect(() => {
-    // dispatch(fetchHotspotsData()) will trigger an update to hotspots
-    // anytime hotspots update, update rewards and following
-    // Separating these calls allows the UI to respond more quickly
-    dispatch(fetchFollowedHotspotsFromBlock())
-    dispatch(fetchRewards())
-  }, [hotspots, dispatch])
+    dispatch(fetchRewards({ fetchType: fleetModeEnabled ? 'followed' : 'all' }))
```
allenan, line 90: is there some state they could get into where they follow too many hotspots and it bogs down too much to where they can't get in to unfollow them? Should there be an upper limit on followed hotspots? at least until we have the API sorted out?

src/store/user/appSlice.ts
```diff
@@ -49,13 +53,15 @@ export const restoreUser = createAsyncThunk<Restore>(
   'app/restoreUser',
   async () => {
     const vals = await Promise.all([
-      getSecureItem('accountBackedUp'),
```
allenan, line 52: you could potentially destructure this like `const [accountBackedUp, requirePin, ...] = await Promise.all([...])` if you don't want to keep track of which index each item is in

src/utils/appDataClient.ts
```diff
@@ -140,6 +140,20 @@ export const getAccount = async (address?: string) => {
...
+  address?: string
+  numDaysBack?: number
+}) => {
+  Logger.breadcrumb('getAccountRewards', breadcrumbOpts)
+  const accountAddress = opts?.address || (await getAddress())
+  if (!accountAddress) return
+
+  const initialDate = new Date()
+  const endDate = new Date()
+  endDate.setDate(initialDate.getDate() - (opts?.numDaysBack || 1))
```
allenan, line 153: this works, there are also some handy functions in the `date-fns` package for working with stuff like this

allenan (approved):
looks good to me

src/utils/appDataClient.ts
matthewcarlreetz: yeah, that's what I would normally use, but I copied this from explorer 😄

matthewcarlreetz: I'll use date-fns to be consistent with other date math in the app.

src/store/user/appSlice.ts
matthewcarlreetz: yes! That's so much better.

src/features/hotspots/root/HotspotsScreen.tsx
matthewcarlreetz: Yes, we discussed this at one point and decided not to handle it for the moment. I will raise the question again and can address it in a separate PR if needed.

PR #816, opened 2021-07-20
Bugfix/808 sync status

### LINKED ISSUE #808: new sync

matthewcarlreetz:
Closes #808

### DIFF
--- a/src/features/hotspots/details/HotspotDetails.tsx
+++ b/src/features/hotspots/details/HotspotDetails.tsx
@@ -46,8 +46,8 @@ import { useColors, useSpacing } from '../../../theme/themeHooks'
 import HotspotSheetHandle from '../root/HotspotSheetHandle'
 import { hp } from '../../../utils/layout'
 import sleep from '../../../utils/sleep'
-import { fetchSyncStatus } from '../../../store/hotspots/hotspotsSlice'
 import usePrevious from '../../../utils/usePrevious'
+import useHotspotSync from '../useHotspotSync'
 
 const hitSlop = { top: 24, bottom: 24 } as Insets
 
@@ -85,15 +85,7 @@ const HotspotDetails = ({
     useSelector(
       (state: RootState) => state.hotspotDetails.hotspotData[address],
     ) || {}
-  const blockHeight = useSelector(
-    (state: RootState) => state.heliumData.blockHeight,
-  )
-  const syncStatuses = useSelector(
-    (state: RootState) => state.hotspots.syncStatuses,
-  )
-  const hotspotSyncBuffer = useSelector(
-    (state: RootState) => state.features.hotspotSyncBuffer,
-  )
+
   const listRef = useRef<BottomSheet>(null)
   const scrollViewRef = useRef<BottomSheetScrollViewType>(null)
   const [isRelayed, setIsRelayed] = useState(false)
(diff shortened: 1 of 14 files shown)


src/features/hotspots/details/HotspotDetails.tsx
```diff
@@ -164,16 +166,10 @@ const HotspotDetails = ({
...
 
-    dispatch(
-      fetchSyncStatus({
-        address,
-        statusTime: hotspot?.status?.timestamp,
-        blockHeight,
-      }),
-    )
-  }, [address, blockHeight, dispatch, hotspot?.status?.timestamp])
+    dispatch(fetchSyncStatus({ hotspot, hotspotSyncBuffer }))
```
allenan, line 171: feels a little odd for a view to know about sync buffer?

src/features/hotspots/details/HotspotStatusBanner.tsx
```diff
@@ -5,40 +5,63 @@ import React, { useState, useEffect, memo, useMemo, useCallback } from 'react'
...
-  const hotspotSyncStatus = useMemo(() => syncStatuses[hotspot.address], [
-    hotspot.address,
-    syncStatuses,
-  ])
+
+  const syncStatusMessage = useMemo(() => {
+    if (!blockHeight || !syncStatus || hotspot.status?.online !== 'online')
+      return null
+
+    const timeAgo = hotspot.status?.timestamp
```
allenan, line 45: I'm seeing this logic repeat, can it be centralized somewhere? like a `<SyncStatusMessage />`?

src/utils/hotspotUtils.ts
```diff
@@ -1,40 +1,36 @@
...
+  hotspotBlockHeight: number
+  blockHeight?: number
+  hotspotSyncBuffer?: number
+}): HotspotSyncStatus => {
+  if (!blockHeight || !hotspotSyncBuffer) {
+    return 'none'
+  }
 
   const syncedRatio = hotspotBlockHeight / blockHeight
   const percentSynced = round(syncedRatio * 100, 2)
```
allenan, line 20: I don't think the concept of percent synced really applies anymore. It's more like hotspot was synced/syncing/offline as of x time ago.

allenan: I guess maybe diagnostics could use the more traditional sync logic, but maybe that should be separate from the redux/gossip sync logic?

allenan:
the logic throughout here feels pretty difficult to parse, guessing just how it's evolved. But it's pretty difficult to trace a line through what's happening and the function signature for `getSyncStatus` for example requires the caller to know a lot about the internal state of sync logic

src/features/hotspots/details/HotspotDetails.tsx
matthewcarlreetz: I agree. I guess I could pull it inside the slice from state. That's probably better.

matthewcarlreetz:
> the logic throughout here feels pretty difficult to parse, guessing just how it's evolved. But it's pretty difficult to trace a line through what's happening and the function signature for `getSyncStatus` for example requires the caller to know a lot about the internal state of sync logic

Yeah, you're right. I'll try to refactor it for clarity.

allenan (approved):
nice refactor

PR #821, opened 2021-07-22
Use wallet api to fetch hotspot rewards.

### LINKED ISSUE #797: Optimize Hotspots List Loading
The hotspot listview makes a call to `hotspot/rewards/sum` for each item in the list. This n+1 operation becomes a burden for accounts with 20+ hotspots. Moving these requests to the wallet api will reduce the burden on the app.

- [x] Create Wallet API endpoint that will accept a list of hotspot addresses and return their rewards
- [x] Cache those rewards in wallet api db for future requests. Update a record if it is older than x minutes. Not sure if this is necessary, but would reduce throughput.
- [ ] Update app to use new wallet api endpoint for rewards

matthewcarlreetz:
Closes #797


src/store/hotspots/hotspotsSlice.ts
```diff
@@ -30,9 +38,8 @@ export type HotspotsSliceState = {
   followedHotspotsObj: Record<string, Hotspot>
   followedHotspots: Hotspot[]
   order: HotspotSort
-  rewards?: Rewards
+  rewards: Rewards
```
tyler-whitman, line 41: Would it be beneficial to use a cache record here like we do with the accountRewards? I guess its also caching in wallet API?

tyler-whitman (approved):
Change looks good

src/store/hotspots/hotspotsSlice.ts
matthewcarlreetz: You're probably right, but it's only one api call at a time and there's caching on the wallet api, so for now I say we just leave it.

PR #826, opened 2021-07-23
Fix infinite loop when deep linking to hotspot.

matthewcarlreetz:
`helium://hotspot/:address`

### DIFF
--- a/src/features/hotspots/root/HotspotsView.tsx
+++ b/src/features/hotspots/root/HotspotsView.tsx
@@ -282,7 +282,8 @@ const HotspotsView = ({
     }
 
     fetchHotspot()
-  }, [dispatch, handlePresentDetails, params])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [params])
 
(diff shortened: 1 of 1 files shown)


src/features/hotspots/root/HotspotsView.tsx
```diff
@@ -282,7 +282,8 @@ const HotspotsView = ({
     }
 
     fetchHotspot()
-  }, [dispatch, handlePresentDetails, params])
```
tyler-whitman, line 285: This used to work right? What changed?

matthewcarlreetz: I think handlePresentDetail now has a dependency that causes it to change when you deep link

matthewcarlreetz: which causes the useEffect to run again

tyler-whitman: Ah ok

tyler-whitman (approved):

PR #830, opened 2021-07-23
Update welcome message to include validators.

matthewcarlreetz:


src/features/hotspots/root/WelcomeOverview.tsx
```diff
@@ -38,19 +41,33 @@ const TimeOfDayTitle = ({ date }: { date: Date }) => {
...
+  )
+
+  const loading = useMemo(() => accountRewards.loading || validatorsLoading, [
     accountRewards.loading,
+    validatorsLoading,
   ])
 
+  useVisible({
+    onAppear: () => {
+      dispatch(fetchValidators())
```
tyler-whitman, line 67: Should we move this to hotspots view where the `fetchAccountRewards()` happens? Or maybe move that here?

src/utils/appDataClient.ts
```diff
@@ -52,6 +52,15 @@ export const getHotspots = async () => {
   return newHotspotList.takeJSON(1000)
 }
 
+export const getValidators = async () => {
+  Logger.breadcrumb('getValidators', breadcrumbOpts)
+  const address = await getAddress()
+  if (!address) return []
+
+  const newValidatorsList = await client.account(address).validators.list()
+  return newValidatorsList.takeJSON(1000)
```
tyler-whitman, line 61: Does this api call have a cursor? If not maybe we should use the `MAX` variable instead to make it more clear?

tyler-whitman (approved):
Looks good, tested it with a no validator but hotspot account and an only validator account. Don't have one with both to try.

src/features/hotspots/root/WelcomeOverview.tsx
matthewcarlreetz: yes, good call

src/utils/appDataClient.ts
matthewcarlreetz: I changed it to max and a few others that should be using too.

Issue #135, opened 2021-02-01
Sender assurance for QR payments


rawrmaan:
Add an optional field to the JSON for `payment` type QR scans: `sender_address`. This will be useful for large operations who regularly scan QR codes for payments from multiple accounts. It will allow them to avoid the unfortunate and irreversible mistake of sending from the wrong wallet.

PR #831, opened 2021-07-24
[#135] Sender assurance for QR payments

MichaelBrew:
Feature request: https://github.com/helium/hotspot-app/issues/135

## Summary
Adds support for a "sender_address" field from scanned QR code JSON payloads when sending payments. If a "sender_address" is provided, that address must match the account address stored in the "address" secure item on device. Otherwise, an error dialog is presented on the ScanScreen describing the current/sender address mismatch, and leaves the user on the ScanScreen. Similarly, If the provided sender address isn't even a valid wallet address, an error dialog is presented and leaves the user on the SendScreen.

The "sender_address" field is supported in all expected payment QR code JSON formats:
```
{ type, senderAddress?, address, amount?, memo? }
{ type, senderAddress?, payees: {[payeeAddress]: amount} }
{ type, senderAddress?, payees: {[payeeAddress]: { amount, memo? }} }
```

## Screenshots
| Invalid Sender Address | Mismatched Sender Address
| ---- | ---- |
| Snip20210731_37 | Snip20210731_38 |

## Implementation
This PR simply updates `parseBarCodeData` and `AppLinkPayment` to support an optional `senderAddress` field. If a valid senderAddress is provided, we verify that it matches the current wallet address by fetching from `secureAccount.getAddress()` and comparing with `_.isEqual`. If they don't match (or sender address is simply invalid), we throw custom `MismatchedAddressError` or `InvalidAddressError` errors respectively which are caught and handled by ScanView (which renders appropriate messaging using `showOKAlert` from the `useAlert` hook).

## i18n
The following strings have been added and translated to Japanese, Korean, and simplified Chinese (with the help of Google Translate):
- `send.scan.parse_code_error`: 'Unable to parse QR code'
- `send.scan.invalid_sender_address`: 'Sender Address in QR code is not a valid wallet account address.'
- `send.scan.mismatched_sender_address`: 'Sender Address in QR code does not match wallet account address. Addresses must match in order to proceed.'

## Testing
These changes have thus far been tested by overriding the "X" button on the ScanView to directly call `handleBarCodeScanned({ ... })` with varying JSON payloads.
- ✅  The 3 existing payment JSON formats should work as expected without `senderAddress`
- ✅  The 3 existing payment JSON formats should work as expected when `senderAddress` is defined as the current address stored on device
- ✅  The 3 existing payment JSON formats should display an error dialog when `senderAddress` is defined as an address that does _not_ match current address stored on device
- ✅  The 3 existing payment JSON formats should display an error dialog when `senderAddress` is an invalid wallet address
- 🟠  `HotspotSetupExternalScreen.tsx` still works as expected
  - Any help / advice testing this would be greatly appreciated (want to ensure that declaring the inline callback assigned to `handleBarCodeScanned` as async doesn't break anything)

### DIFF
--- a/src/providers/AppLinkProvider.tsx
+++ b/src/providers/AppLinkProvider.tsx
@@ -128,9 +128,9 @@ const useAppLink = () => {
    * The data scanned from the QR code is expected to be one of these possibilities:
    * (1) A helium deeplink URL
    * (2) address string
-   * (3) stringified JSON object { type, address, amount?, memo? }
-   * (4) stringified JSON object { type, payees: {[payeeAddress]: amount} }
-   * (5) stringified JSON object { type, payees: {[payeeAddress]: { amount, memo? }} }
+   * (3) stringified JSON object { type, senderAddress?, address, amount?, memo? }
+   * (4) stringified JSON object { type, senderAddress?, payees: {[payeeAddress]: amount} }
+   * (5) stringified JSON object { type, senderAddress?, payees: {[payeeAddress]: { amount, memo? }} }
    */
   const parseBarCodeData = useCallback(
     (data: string, scanType: AppLinkCategoryType): AppLink | AppLinkPayment => {
@@ -179,9 +179,10 @@ const useAppLink = () => {
         if (type === 'payment') {
           let scanResult: AppLinkPayment
           if (rawScanResult.address) {
-            // Case (3) stringified JSON { type, address, amount?, memo? }
+            // Case (3) stringified JSON { type, senderAddress?, address, amount?, memo? }
             scanResult = {
               type,
+              senderAddress: rawScanResult.senderAddress,
               payees: [
                 {
                   address: rawScanResult.address,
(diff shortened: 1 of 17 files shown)


cokes518:
I wonder if we can avoid showing this payment screen all together and just show a popup after scanning the QR code that says something like:

"Unable to parse QR code"
"Sender Address in QR code does not match wallet account address. Addresses must match in order to proceed."
"OK"

src/providers/AppLinkProvider.tsx
```diff
@@ -25,6 +28,48 @@ import {
...
+  }
+}
+const assertValidAddress = (address: string, addressType: AddressType) => {
+  if (!address) throw new MissingAddressError(addressType)
+  if (!Address.isValid(address)) throw new InvalidAddressError(addressType)
+}
+const assertCurrentSenderAddress = async (senderAddress: string) => {
+  const current = await getAddress()
+  const sender = Address.fromB58(`${senderAddress}`)
+  if (!isEqual(current, sender)) {
```
tyler-whitman, line 68: instead of doing a deep object compare here, you could just use `current?.b58 !== senderAddress`

tyler-whitman (commented):
Code looks good, just the one comment. Also you are going to get conflicts when we merge your 3 open pull requests.

tyler-whitman:
I made the changes in https://github.com/helium/hotspot-app/pull/955

tyler-whitman:
merged in #955

PR #833, opened 2021-07-26
Make iPad build fullscreen

compojoom:
This change allows the app to run fullscreen on iPad instead of a scaled down window.

Before:
Screenshot 2021-07-26 at 10 25 36

after:
grafik

I quickly checked other screns in the app and they looked fine.


tyler-whitman (changes requested):
Thanks for the contribution this looks good. I tested as well and the only thing I noticed is the ticker in the `HotspotsTicker` component is not working correctly because it isnt long enough to reach the ends of the screen anymore. It's supposed to animate scroll.

You can change `HotspotsTicker.tsx` line 86 to:
`{isTablet() ? text + text : text}`

This doubles the text for tablets so the text reaches and it works again.

compojoom:
@tyler-whitman Good catch 👍 . I just added the change - thanks for making it idiot easy to apply the fix :)

compojoom:
@tyler-whitman good now?

PR #838, opened 2021-07-27
Add RisingHF Outdoor Hotspot

lingtaoluorxhf:
This is to add the RisingHF Outdoor Hotspot to the hotspot app. The only thing that needs changing is the onboarding ID in src/makers/risinghf/index.ts, which I'll be in talks with DeWi for. Please let me know if any changes are needed.

### DIFF
--- a/src/makers/index.ts
+++ b/src/makers/index.ts
@@ -5,6 +5,7 @@ import helium from './helium'
 import longAP from './longAP'
 import nebra from './nebra'
 import rak from './rak'
+import risinghf from './risinghf'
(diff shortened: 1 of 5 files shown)


cokes518 (commented):
+1 from me. We'll revisit this PR when RisingHF passes Hardware Audit.

cokes518:
we'll merge this once MOC approves, a maker account is created, and RisingHF is assigned an `id`; in that order.

cokes518:
please update `id` with your assigned number

PR #839, opened 2021-07-27
Handle data only hotspots.

### LINKED ISSUE #815: Support Data-Only Assert Locations
Data-Only Hotspots can be added on chain via CLI. These Hotspots also appear in the app and if users assert a data-only Hotspot, the transaction will fail because it assumes the wrong fees.

For Data-Only Hotspots, the assert location fee is $5

### LINKED ISSUE #822: Add UI Support for Data-Only Hotspots
Data Only Hotspots are a subset of regular Hotspots in that they can't PoC, do not affect transmit scale, do not sync, and have no witnesses.

We can identify them in the Hotspot json as `mode: dataonly`.

matthewcarlreetz:
Handle data only in hotspot detail view.
Set data only hotspot reassert fee to chain var.
Remove update antenna for dataonly hotspots.

Closes #822
Closes #815


src/utils/assertLocationUtils.tsx
```diff
@@ -70,26 +71,45 @@ export const assertLocationTxn = async (
...
+  dataOnly = false,
+}: {
+  nonce?: number
+  accountIntegerBalance?: number
+  onboardingRecord?: OnboardingRecord
+  dataOnly?: boolean
+}) => {
+  let isFree = false
+  if (!dataOnly) {
+    isFree = await hasFreeLocationAssert(nonce, onboardingRecord)
```
tyler-whitman, line 87: Should we move this `dataOnly` check to `hasFreeLocationAssert` maybe? I see in the `assertLocationTxn` line 36 is also checking this, does it need a check for dataOnly there too? Maybe in `hasFreeLocationAssert` we return false if its dataOnly?

tyler-whitman (changes requested):
Do you think we should also hide the Earnings / Progress / Witnesses tabs? At least Earnings and Witnesses will always be the default state. I guess some of the Progress will complete around data credits but still not an ideal experience.

cokes518:
> Do you think we should also hide the Earnings / Progress / Witnesses tabs? At least Earnings and Witnesses will always be the default state. I guess some of the Progress will complete around data credits but still not an ideal experience.

image

in the mock ups, I've hidden the progress and witnesses tab, and they will have earnings eventually. 

progress is not relevant since the only interesting "checklist" item is "data transfer", and that will be obvious once they see earnings, or they can look on explorer.

src/utils/assertLocationUtils.tsx
matthewcarlreetz: My assumption is that `dataOnly` hotspots won't ever have a free assert. Is that right @cokes518 ?

cokes518: correct. no free asserts.

src/utils/assertLocationUtils.tsx
```diff
@@ -41,9 +55,16 @@ export const assertLocationTxn = async (
   }
 
   const antennaGain = decimalGain * 10
-  const stakingFee = updatingLocation
-    ? Transaction.stakingFeeTxnAssertLocationV1
-    : 0
+  let stakingFee = 0
+  if (updatingLocation) {
+    if (dataOnly) {
```
matthewcarlreetz, line 60: Does this look right @tyler-whitman ?

tyler-whitman: Yeah looks good, just realized that `hasFreeLocationAssert` doesn't return a promise though so we can remove the `await`

tyler-whitman (approved):
Looks good, thanks for making those changes

PR #883, opened 2021-08-12
Assert crash and diagnostic report update

### LINKED ISSUE #877: Add instructions to email generated from diagnostics
Above the output of the diagnostic report, we should add some instructions to add more context for support teams who are on the receiving end.

Once the email is generated:

**Please add more details to the issue you're experiencing below**

Hotspot: <name>
Hotspot Maker: <name>
etc.

### LINKED ISSUE #878: Set Location fails if a user skips assert in onboarding flow
We've identified the crash, just need to fix it.

Steps to repro:
1. onboard a hotspot, skip assert location in the onboarding flow
2. go to settings > remote assert
3. crash

tyler-whitman:
- Fix crash when updating hotspot location for hotspots that skipped assert during onboarding
- Add more detail to the diagnostic report email

closes #878 
closes #877

### DIFF
--- a/src/features/hotspots/settings/sendReport.ts
+++ b/src/features/hotspots/settings/sendReport.ts
@@ -45,6 +45,7 @@ export default async ({
   let url = `mailto:${supportEmail}`
 
   const body = [
+    "**Please add more details to the issue you're experiencing below**\n\n",
     `Hotspot: ${kebabCase(animalHash(gateway))}`,
     `Hotspot Maker: ${hotspotMaker}`,
     `Address: ${gateway}`,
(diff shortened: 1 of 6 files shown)


src/features/hotspots/settings/sendReport.ts
```diff
@@ -45,6 +45,7 @@ export default async ({
   let url = `mailto:${supportEmail}`
 
   const body = [
+    "**Please add more details to the issue you're experiencing below**\n\n",
```
cokes518, line 48: I think you need to i18n this and put the string in the en.ts file.

tyler-whitman: Was wondering about that, figured we might want to keep support emails in english? But I can move to translations if you prefer.

allenan (approved):

PR #893, opened 2021-08-17
check to make sure map coords are a number

### LINKED ISSUE #867: Selecting Location tab in Hotspot Settings fails when there is no asserted location
Steps to reproduce:
1. Add a Hotspot using the onboarding flow
2. Set a location later
3. go to Hotspot details, tap settings
4. Tap Update Settings
5. select Update Location tab
6. crash

Has happened to several users on Android, iOS but the devs are unable to reproduce. Will wait for more crash logs to come through to see if we can dig deeper.

tyler-whitman:
hopefully fixes #867

### DIFF
--- a/src/components/Map.tsx
+++ b/src/components/Map.tsx
@@ -19,6 +19,7 @@ import { StyleProp, ViewStyle } from 'react-native'
 import { useTranslation } from 'react-i18next'
 import { h3ToGeo } from 'h3-js'
 import Config from 'react-native-config'
+import { isNumber } from 'lodash'
 import Box from './Box'
 import Text from './Text'
 import NoLocation from '../assets/images/no-location.svg'
@@ -102,7 +103,12 @@ const Map = ({
   }, [onMapMoved])
 
   const centerUserLocation = useCallback(() => {
-    const hasCoords = userCoords && userCoords.longitude && userCoords.latitude
+    const hasCoords =
+      userCoords &&
+      userCoords.longitude &&
+      userCoords.latitude &&
+      isNumber(userCoords.longitude) &&
+      isNumber(userCoords.latitude)
     camera.current?.setCamera({
       centerCoordinate: hasCoords
         ? [userCoords.longitude, userCoords.latitude]
@@ -125,7 +131,13 @@ const Map = ({
   )
 
   useEffect(() => {
-    if (!showUserLocation || !userCoords.latitude || !userCoords.longitude)
+    if (
+      !showUserLocation ||
+      !userCoords.latitude ||
+      !userCoords.longitude ||
+      !isNumber(userCoords.latitude) ||
+      !isNumber(userCoords.longitude)
+    )
       return
 
     camera.current?.setCamera({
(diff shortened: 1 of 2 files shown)


src/utils/mapUtils.ts
```diff
@@ -27,7 +28,7 @@ export const findBounds = (
   coords: number[][],
   paddingBottom = 250,
 ): MapBounds | undefined => {
-  if (coords.length === 0) {
+  if (coords.length === 0 || !isNumber(coords[0]) || !isNumber(coords[1])) {
```
matthewcarlreetz, line 31: Wouldn't both `coords[0]` and `coords[1]` be arrays here?

src/components/Map.tsx
```diff
@@ -102,7 +103,12 @@ const Map = ({
   }, [onMapMoved])
 
   const centerUserLocation = useCallback(() => {
-    const hasCoords = userCoords && userCoords.longitude && userCoords.latitude
+    const hasCoords =
```
matthewcarlreetz, line 106: Null and undefined aren't numbers so I think we can shorten this to 
```ts
 const hasCoords =
      userCoords &&
      isNumber(userCoords.longitude) &&
      isNumber(userCoords.latitude
```

src/components/Map.tsx
```diff
@@ -125,7 +131,13 @@ const Map = ({
   )
 
   useEffect(() => {
-    if (!showUserLocation || !userCoords.latitude || !userCoords.longitude)
+    if (
```
matthewcarlreetz, line 134: I think we could do the same here...
```ts
   if (
      !showUserLocation ||
      !isNumber(userCoords.latitude) ||
      !isNumber(userCoords.longitude)
    )
      return
```

src/components/Map.tsx
```diff
@@ -195,13 +207,19 @@ const Map = ({
     return findBounds(boundsLocations, cameraBottomOffset)
   }, [mapCenter, cameraBottomOffset, selectedHex, selectedHotspot, witnesses])
 
-  const defaultCameraSettings = useMemo(
-    () => ({
+  const defaultCameraSettings = useMemo(() => {
+    const centerCoordinate =
```
matthewcarlreetz, line 211:
```ts
    const centerCoordinate =
      mapCenter?.length === 2 &&
      isNumber(mapCenter[0]) &&
      isNumber(mapCenter[1])
        ? mapCenter
        : defaultLngLat
```

src/utils/mapUtils.ts
tyler-whitman: Good catch, moved the check down to check the actual numbers

src/components/Map.tsx
tyler-whitman: updated

src/components/Map.tsx
tyler-whitman: updated

src/components/Map.tsx
tyler-whitman: updated

matthewcarlreetz (approved):

PR #898, opened 2021-08-19
Feature/account settings

matthewcarlreetz:
The app-level settings `isFleetModeEnabled`, `hasFleetModeAutoEnabled`, and `convertHntToCurrency` all have been moved to account settings (which is backed by the wallet API). Any app-level settings will now live in `appSlice` and account settings will live in `accountSlice`.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,7 +16,7 @@ import { NavigationContainer } from '@react-navigation/native'
 import { theme } from './theme/theme'
 import NavigationRoot from './navigation/NavigationRoot'
 import { useAppDispatch } from './store/store'
-import appSlice, { restoreUser } from './store/user/appSlice'
+import appSlice, { restoreAppSettings } from './store/user/appSlice'
 import { RootState } from './store/rootReducer'
 import { fetchData } from './store/account/accountSlice'
 import BluetoothProvider from './providers/BluetoothProvider'
@@ -36,6 +36,8 @@ import notificationSlice, {
 } from './store/notifications/notificationSlice'
 import AppLinkProvider from './providers/AppLinkProvider'
 import { navigationRef } from './navigation/navigator'
+import useSettingsRestore from './utils/useAccountSettings'
+import useMount from './utils/useMount'
 
 SplashScreen.preventAutoHideAsync().catch(() => {
   /* reloading the app might trigger some race conditions, ignore them */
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -130,21 +135,19 @@ const App = () => {
...
 
-  // restore user and then fetch initial data
+  // update initial data when account is restored or app comes into foreground from background
   useEffect(() => {
-    if (!isRestored) {
-      dispatch(restoreUser())
-    } else {
+    if (
+      isRestored ||
+      (prevAppState === 'background' && appState === 'active')
```
tyler-whitman, line 142: Why do we need to call `loadInitialData` even when `isRestored` is false? Is there a timing issue? That means they would be logged out but we still make the call right?

src/features/moreTab/more/MoreScreen.tsx
```diff
@@ -52,6 +56,14 @@ const MoreScreen = () => {
   const navigation = useNavigation<MoreNavigationProp & RootNavigationProp>()
   const spacing = useSpacing()
 
+  useEffect(
+    () =>
+      navigation.addListener('focus', () => {
+        dispatch(fetchAccountSettings())
```
tyler-whitman, line 62: This also gets called on app launch right? Is it needed here in case the settings change after it launches? Maybe on another app or something?

src/store/helium/heliumDataSlice.ts
```diff
@@ -66,12 +65,7 @@ export const fetchInitialData = createAsyncThunk<HeliumDataState>(
 const heliumDataSlice = createSlice({
   name: 'heliumData',
   initialState,
-  reducers: {
-    signOut: () => {
-      signOut()
-      return { ...initialState, isRestored: true }
```
tyler-whitman, line 72: So we save this data because its not related to an account?

tyler-whitman (approved):
Looks good, just a couple questions for you

src/App.tsx
matthewcarlreetz: I'm wondering if we even care about `isRestored` at all here. The only thing that is dependent on the account is notifications and I don't think we need to load them here anyway. I'll take a look at refactoring this...

src/features/moreTab/more/MoreScreen.tsx
matthewcarlreetz: yeah, it's to keep it updated if settings change from a different device.

src/store/helium/heliumDataSlice.ts
matthewcarlreetz: yeah, there's no reason to reset it. It's all account agnostic.

PR #899, opened 2021-08-19
downgrade sentry version

tyler-whitman:
Sentry version 2.6.2 required a new target OS version for iOS, so we will downgrade it back to 2.2.0 for now.


tyler-whitman:
Error was incorrect, just needed a refresh of pod lockfile fixed in https://github.com/helium/hotspot-app/pull/900

PR #923, opened 2021-09-01
Hide hotspot fixes

tyler-whitman:
- resolves issues with un-hiding your last hidden hotspot
- adds support to hide more than 5 hotspots

We now store hotspots as json instead of a comma separated list

### DIFF
--- a/src/store/account/accountSlice.ts
+++ b/src/store/account/accountSlice.ts
@@ -271,7 +271,7 @@ const accountSlice = createSlice({
       },
     )
     builder.addCase(
-      updateSetting.fulfilled,
+      updateSetting.pending,
       (
         state,
         {
--- a/src/store/account/accountSlice.ts
+++ b/src/store/account/accountSlice.ts
@@ -271,7 +271,7 @@ const accountSlice = createSlice({
       },
     )
     builder.addCase(
-      updateSetting.pending,
+      updateSetting.fulfilled,
       (
         state,
         {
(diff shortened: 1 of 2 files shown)


src/store/account/accountSlice.ts
```diff
@@ -271,7 +271,7 @@ const accountSlice = createSlice({
       },
     )
     builder.addCase(
-      updateSetting.pending,
+      updateSetting.fulfilled,
```
tyler-whitman, line 274: @matthewcarlreetz I'm not sure we needed this to be pending? But if it is, then when the API call fails the app data still changes.

matthewcarlreetz: I did it that way so that the ui could update without having to wait for the api response. Maybe we just handle failed instead of changing this to fulfilled?

tyler-whitman: If we added and then removed it might feel weird to see the UI change and then change back when it fails? For me even with it at fulfilled it loaded quickly. I guess if the api is slow it might lag a second. What do you think, I'm leaning towards leaving it as fulfilled?

matthewcarlreetz: My feeling is the immediate feedback is needed in the case that the api is slow. In the rare case that the api is down, I think it's ok to change the ui back if the call fails.

matthewcarlreetz (approved):

PR #946, opened 2021-09-20
Feature/934 validators filters

matthewcarlreetz:


src/features/hotspots/root/HotspotsList.tsx
```diff
@@ -123,22 +291,49 @@ const HotspotsList = ({
...
+            showCarot
+            loading={loadingHotspotRewards}
+            totalReward={hotspotRewards[item.address]}
+            hidden={hiddenAddresses?.includes(item.address)}
+          />
+        )
+      }
+      if (isValidator(item)) {
+        return (
+          <ElectedValidatorItem
```
tyler-whitman, line 321: Maybe we refactor this to `ValidatorListItem`?

matthewcarlreetz: that sounds good

tyler-whitman (approved):
Code looks good and working well for me, only other thing I noticed is that the HNT amount for the validator list item doesn't convert to local currency when using the toggle. Thats the way it works it is on the consensus list too though.

matthewcarlreetz:
> Code looks good and working well for me, only other thing I noticed is that the HNT amount for the validator list item doesn't convert to local currency when using the toggle. Thats the way it works it is on the consensus list too though.

oh, good catch. I'll fix that.

tyler-whitman:
> > Code looks good and working well for me, only other thing I noticed is that the HNT amount for the validator list item doesn't convert to local currency when using the toggle. Thats the way it works it is on the consensus list too though.
> 
> oh, good catch. I'll fix that.

Looks good to merge when you are ready!

PR #958, opened 2021-09-29
Add maker merryIoT

browanofficial:
Add maker merryIoT


src/makers/merryIoT/index.ts
```diff
@@ -0,0 +1,9 @@
+import antennas from './antennas'
+import hotspots from './hotspots'
+
+export default {
+  antennas,
+  hotspots,
+  id: 22,
```
cokes518, line 7: this `id` will need to be updated with your maker `id`. you're not guaranteed to get 22.

PR #961, opened 2021-09-30
Feature/network selection

matthewcarlreetz:
Add network selector to settings.
Use selected network option to configure helium js and wallet client.
Always read pending txns from mainnet.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,7 +22,7 @@ import { fetchData } from './store/account/accountSlice'
 import BluetoothProvider from './providers/BluetoothProvider'
 import ConnectedHotspotProvider from './providers/ConnectedHotspotProvider'
 import * as Logger from './utils/logger'
-import { configChainVars } from './utils/appDataClient'
+import { configChainVars, updateClient } from './utils/appDataClient'
 import {
   fetchBlockHeight,
   fetchInitialData,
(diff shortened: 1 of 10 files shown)


src/utils/appDataClient.ts
```diff
@@ -24,7 +25,27 @@ import { fromNow } from './timeUtils'
...
+    (network === 'helium' && client.network === Network.production)
+  )
+}
+
+export const updateClient = (nextNetwork: string) => {
+  const isSame = compareNetwork(nextNetwork)
+  if (!isSame) {
+    const network =
+      nextNetwork === 'helium' ? Network.production : Network.stakejoy
+    client = new Client(network)
```
tyler-whitman, line 45: Could this get reinitialized while things are already partly paged and throw off the cursor? Or do we just call this and everything in the app refreshes?

src/App.tsx
```diff
@@ -86,9 +87,14 @@ const App = () => {
 
   useMount(() => {
     dispatch(restoreAppSettings())
+  })
+
+  useEffect(() => {
+    if (!settingsLoaded) return
```
tyler-whitman, line 93: Does fetchAccountSettings only get dispatched on the more screen? Should we also call it here to makes sure it loads and doesnt block the app?

tyler-whitman (approved):

src/utils/appDataClient.ts
matthewcarlreetz: I had the same concern. A couple of lines down I re-init the txn fetchers and after thorough testing, it all seems to work. Could you test when you have a chance? The wallet API is updated now.

src/App.tsx
matthewcarlreetz: It gets called in App.tsx by the `useAccountSettings` hook as well.

PR #963, opened 2021-10-04
Create currency type selector.

### LINKED ISSUE #848: Add fiat currency selector in settings
**Describe the bug**
Currently, the app is using the phone's locale/country to choose which currency is displayed when the Convert HNT to Currency option is active.
This is a bad behavior because most of the time we cannot change the whole phone's locale to have a different currency in the app.
Ex: we want to display the value in USD but are in Canada, it would display in CAD. We cannot change the phone's locale otherwise the AppStore would not operate in the right country.

**To Reproduce**

1. Go to Settings
2. Scroll down and activate Convert HNT to Currency
3. Notice that the currency is displayed in your phone's locale/country (which is not US) when you want it to be seen in USD.

**Expected behavior**
Have a list of currencies to convert to instead of using the phone's locale/country.

**Smartphone (please complete the following information):**

- Device: iPhone 12
- OS: iOS 14.6
- App Version 3.4.3

**Additional context**
Tag: improvement more than a bug.
I saw a lot of people asking for this feature in a few Discord servers.

matthewcarlreetz:
closes #848

### DIFF
--- a/src/components/HeliumActionSheet.tsx
+++ b/src/components/HeliumActionSheet.tsx
@@ -87,7 +87,7 @@ const HeliumActionSheet = ({
       return initialValue
     }
     const item = data.find((d) => d.value === selectedValue)
-    return item?.label || ''
+    return item?.labelShort || item?.label || ''
   }, [data, initialValue, selectedValue])
 
   const selected = useCallback(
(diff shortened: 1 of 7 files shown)


tyler-whitman (approved):
Looks good and working as expected!

PR #974, opened 2021-10-15
transfer hotspot v2

### LINKED ISSUE #938: support for transfer_hotspot_v2
transfer_hotspot_v2 requires 4 fields:

1. owner_address
2. owner_signature
3. gateway_address
4. new_owner_address

The old owner will pay the transaction fee.

As part of this simplification, the following checks/required fields are removed on the blockchain side but should remain on the app side):
* staleness check removed (check in app)
* poc check removed (check in app)
* HNT field removed (not required)

The new flow will be as follows:
1. Seller goes to Hotspot > Tap Settings > Transfer Hotspot (unchanged)
2. Type Hotspot name to confirm transfer (unchanged)
3. Seller asked to enter a Buyer address. HNT Amount removed (new)
4. PoC liveness check occurs by querying the API of the last successful PoC challenge (if it's not a data-only hotspot). No checks required for data-only hotspots. if the last poc challenge was within the last 24 hours, allow user to proceed with transaction. if last poc challenge was more than 24 hours ago, do not allow user to proceed and disable the button. 
4. Seller presses button to Transfer Hotspot (UI change)
5. Below the button explains that the transaction will just go through, buyer acceptance not required (UI change)
6. Once the transaction clears, both the Buyer and Seller will receive an app notification saying the transaction cleared. (new)

This new flow removes dependencies on wallet-api and should make the feature more streamlined and less error-prone.

For screens, go to Figma > 3.4.0 > transfer_hotspot_v2.

Relates to https://github.com/helium/blockchain-core/issues/978

tyler-whitman:
Adds support for transfer hotspot v2. We still keep support in the app to complete and cancel transfer v1s for any pending, but all newly submitted transfers will now use v2 which no longer needs a buyer signature.

Transfer v2 is not active yet and this still needs to be tested.

closes #938


smarthousetips:
Any update on this? Can’t transfer a hotspot. Button remains greyed out.

matthewcarlreetz (approved):

Issue #976, opened 2021-10-17
[Feature] Scan to assert antenna gain and elevation


rawrmaan:
With the upcoming PoCv11 rollout, it's advised that all hotspot owners assert their antenna `elevation` and `gain` on-chain. This applies to fleet owners as well. Currently, the best-case (most efficient) way to do this for fleet owners is as follows:

1. Display a QR code with the string `helium://hotspot/[address]` to your deployers
2. Have them scan the QR code to go to the hotspot's page in the app
3. They press the Settings button, then Update Hotspot
4. They choose Custom Antenna from the list and manually enter the hotspot's gain and elevation
5. They press Submit

### Proposal

Add the ability to scan `elevation` and `gain` and tap "I Confirm", completing the entire process in two steps. This could be accomplished in either of the following ways:

- Amend the existing scan-to-assert view and `hotspot_location` scan type to accept `elevation` and `gain` parameters in the QR code, keeping in mind that all 3 of the following scenarios will now be possible and should be accounted for:
-- Scan `hotspotAddress`, `latitude`, `longitude`, `elevation` and `gain` all at once
-- Scan `hotspotAddress`, `latitude` and `longitude` only (this is what's current supported)
-- Scan `hotspotAddress`, `elevation` and `gain` only
- Add a new scan type `hotspot_antenna_info` which accepts `hotspotAddress`, `elevation` and `gain` only and create a new UI to go along with it. Downside of this is that it is technically possible to assert location and antenna info all in one `assert_location_v2` transaction to save time and fees, and this would not support that.

cc @MichaelBrew

PR #977, opened 2021-10-17
[#976] Scan to assert antenna gain and elevation

MichaelBrew:
Feature request: https://github.com/helium/hotspot-app/issues/976

## Summary
This PR adds the ability to scan a QR code (structure defined below) to reassert the gain/elevation of a hotspot's antenna within a single screen. The code can be scanned by tapping the QR symbol within the Wallet View, and will render a pre-filled version of the "Update Antenna" screen from Settings.

**QR Code JSON Payload**
```json
{
  "hotspot_address": "11vAu1u...",
  "gain": 1.3,
  "elevation": 3
}
``` 

| Screenshot | Invalid Address | Demo* |
| ---- | ---- | ---- |
| image | Snip20210821_3 | assert_antenna_demo |

_* To facilitate the demo on local simulator, I hooked up the "X" icon from the Scan View into `handleBarCodeScanned` with a hardcoded JSON payload._

## Implementation
This PR adds a new, lightweight `HotspotAntennaUpdateScreen` component nested within `HotspotsStack` stack navigator (alongside `HotspotsScreen`). The hotspot address, elevation, and gain parsed from the QR code are passed to this screen via navigation props. It then renders an updated version of `UpdateHotspotConfig` which accepts three new props:
```javascript
antennaGain?: number
antennaElevation?: number
initState?: State
```
This allows `UpdateHotspotConfig` to be render with pre-filled gain/elevation values. It also pre-fills the antenna maker as "Custom" if a gain is provided so that the gain field can be edited without selecting a specific antenna maker.

It also updates the relationship between `UpdateHotspotConfig` and `HotspotConfigurationPicker` so that the gain and elevation inputs act more as controlled inputs. Now `UpdateHotspotConfig` is responsible for persisting the `gain` and `elevation` numbers, then passes those down to `HotspotConfigurationPicker`, who then passes the values to `TextInput`. However, I maintained the existing behavior where formatting only happens after the user has finished editing (rather than formatting the input upon each character).

## Testing
So far the extent of testing done is basically what's shown in the demo above. The `submitTxn` call has been stubbed out, 
 relying on the assumption that `submitTxn` within the usual "Settings > Update Antenna" flow works as expected.

Have also tested that providing either an invalid address in the QR code payload or a hotspot address that doesn't belong to the currently authenticated user will result in the "error" dialog shown above.

### DIFF
--- a/src/features/hotspots/root/HotspotAntennaUpdateScreen.tsx
+++ b/src/features/hotspots/root/HotspotAntennaUpdateScreen.tsx
@@ -65,7 +65,7 @@ function HotspotAntennaUpdateScreen({ route }: Props) {
                   antennaElevation={elevation}
                   antennaGain={gain}
                   hotspot={hotspot}
-                  initState="antenna"
+                  initState="confirm"
                   onClose={onClose}
                   onCloseSettings={onClose}
                 />
--- a/src/features/hotspots/settings/updateHotspot/UpdateHotspotConfig.tsx
+++ b/src/features/hotspots/settings/updateHotspot/UpdateHotspotConfig.tsx
@@ -86,6 +86,24 @@ const UpdateHotspotConfig = ({
     enableBack(onClose)
   }, [enableBack, onClose])
 
+  useEffect(() => {
+    if (!!(antennaGain || antennaElevation) && initState === 'confirm') {
+      updateLocationFeeForUpdatingAntenna()
+      setIsLocationChange(false)
+    }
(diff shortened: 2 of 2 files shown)


matthewcarlreetz:
Hi Michael. I rebased your PR with the latest changes and pushed it here https://github.com/helium/hotspot-app/tree/scan_antenna_rebased 

I tested with the QR scanner and JSON format you provided, but it doesn't appear to be working. Could you provide a working QR code?

PR #1004, opened 2021-10-27
Update risinghf-out.svg

lingtaoluorxhf:
Because the image size of the previous "risinghf-out.svg" was not coordinated, it did not display well on the app, so we optimized the image size this time.


cokes518:
please review when you can @tyler-whitman @matthewcarlreetz

matthewcarlreetz (approved):
Looks good

PR #1039, opened 2021-11-04
add loading ui for hex selection

### LINKED ISSUE #1045: Hotspot Details > Status pill shows syncing then online
On the latest TestFlight 3.8.0, navigate to an owned Hotspot.

The hotspot status pill shows "syncing" for about 8s until it updates to say "online". If this is the app querying the API for status, then we should show a loading animation and not "syncing". Most folks don't stay on that screen long enough for it to change to "online".

tyler-whitman:
Changes the hex selection interaction to happen instantly and no longer wait for the `/hotspots/hex/id` call to finish. Because of this I added a loading state to the card itself. I also made it so that while a hex is loading you cant tap to load another, because when things are slow they all pile up and then sporadically load all at once.

closes #1045

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -27,7 +27,6 @@
     "jest/no-export": "off",
     "import/prefer-default-export": "off",
     "no-use-before-define": "off",
-    "no-console": [2, { "allow": ["warn", "error"] }],
     "@typescript-eslint/no-use-before-define": "off",
     "default-case": "off",
     "quotes": [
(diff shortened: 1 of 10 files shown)


allenan (approved):
good changes, might want a review from @matthewcarlreetz on this one as well since it hits quite a few places

matthewcarlreetz (commented):
These changes look great, really like the loading animation there. There still is a lot of lag when clicking on the map. I'm seeing it take up to 15 seconds to switch hexes.

matthewcarlreetz:
There's a 300ms to 1.5s delay between when you tap on a hex and when `onPress` gets called on the vector source. I believe it happens because querying the map to find the feature hex id is an expensive time-consuming operation. I'm going to create a proposal on how to fix this.

matthewcarlreetz:
Check this out and let me know what you think
https://github.com/helium/hotspot-app/pull/1043

PR #1041, opened 2021-11-04
disable console for release builds

tyler-whitman:
There are some new warning logs coming from react native for an EventEmitter deprecated call. Many libraries use this call so it gets a bunch of deprecation warning logs.

For this reason I disabled logs in release builds, and we can update they libraries causing the logs when they have a fix.


cokes518 (approved):
PM +1

PR #1059, opened 2021-11-08
Fix Android top crashes

### LINKED ISSUE #1055: App crash when pairing
on android, 3.8.0. user reporting the app is crashing.

wallet address: https://discord.com/channels/404106811252408320/733699201887764511/907331821648347157

tyler-whitman:
resolves a couple top crashes on Android

closes #1055

### DIFF
--- a/src/utils/animateTransition.ts
+++ b/src/utils/animateTransition.ts
@@ -19,5 +19,16 @@ export default (
     console.log('animateTransition:', { id, enabledOnAndroid })
   }
 
-  LayoutAnimation.configureNext(config || LayoutAnimation.Presets.easeInEaseOut)
+  LayoutAnimation.configureNext(
+    config || {
+      duration: 300,
+      create: {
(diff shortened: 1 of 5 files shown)


src/utils/animateTransition.ts
```diff
@@ -19,5 +19,16 @@ export default (
     console.log('animateTransition:', { id, enabledOnAndroid })
   }
 
-  LayoutAnimation.configureNext(config || LayoutAnimation.Presets.easeInEaseOut)
+  LayoutAnimation.configureNext(
```
matthewcarlreetz, line 22: Curious why this is necessary?

matthewcarlreetz (approved):

src/utils/animateTransition.ts
tyler-whitman: One of the top crashes we have has to do with removing view layers that dont exist and I think it comes from this layout animation. I found another person online that said this change made it not crash and in the app it looks good animation wise so figured we could try it

PR #1062, opened 2021-11-09
update hotspot location map 

tyler-whitman:
- update the map to use a PointAnnotation instead of an overlay on top of the map to show the location marker so the exact coordinate location is always shown correctly
- add maxFontSizeMultiplier to text that was too big with huge fonts


src/features/hotspots/settings/updateHotspot/HotspotLocationPreview.tsx
```diff
@@ -15,45 +18,89 @@ type Geocode = {
...
-        </Text>
-      </Box>
+        <MapboxGL.Camera
+          defaultSettings={{
+            centerCoordinate: mapCenter || defaultLngLat,
+            zoomLevel: 17,
+          }}
+          maxZoomLevel={17}
+        />
+        <MapboxGL.PointAnnotation
```
matthewcarlreetz, line 94: Nice. This is definitely a better solution than what we had. 👍

matthewcarlreetz (approved):

PR #1079, opened 2021-11-12
add HNTenna option to app

curiousfokker:
Request to add the HNTenna as an antenna option in the app.


cokes518:
superseded by #1080

PR #1081, opened 2021-11-12
add redux persist and app level cache

tyler-whitman:


src/App.tsx
```diff
@@ -82,6 +82,8 @@ const App = () => {
     'expo-permissions is now deprecated',
   ])
 
+  LogBox.ignoreAllLogs()
```
matthewcarlreetz, line 85: We should remove this. Are there lots of errors now regarding non-serializable objects?

matthewcarlreetz (commented):
One thing I noticed is that when you select a filter on the activity tab it will persist on next app launch. I don't think we want that? We might want to blacklist `activitySlice.filter`.

matthewcarlreetz:
I'd like to explore if there are API calls that we can eliminate altogether now that we're caching. I created one example here...
https://github.com/helium/hotspot-app/pull/1082

matthewcarlreetz:
I'm still curious if caching activity in the wallet API would be a good idea

src/App.tsx
tyler-whitman: Whoops forgot to push my commit to remove this. Was just using it for testing with offline network

src/utils/useCurrency.ts
```diff
@@ -140,16 +140,25 @@ const useCurrency = () => {
...
+          let [intStr, decStr] = balance
             .toString(undefined, {
               decimalSeparator,
               groupSeparator,
               showTicker: false,
             })
             .split(decimalSeparator)
 
+          // when there is no network the toString method from helium js may not work properly
+          if (intStr === '[object Object]') {
```
matthewcarlreetz, line 155: should we fix helium-js?

matthewcarlreetz (approved):

src/utils/useCurrency.ts
tyler-whitman: Its not helium-js its that the object that gets persisted doesnt get the toString method properly so it just ends up returning `[object Object]` when its loaded from cache

PR #1090, opened 2021-11-16
fix lag when pressing update hotspot button

### LINKED ISSUE #1093: Remove API / owner look up when pairing with a Hotspot
The app currently looks up (via API) who the owner of a Hotspot is before pairing. This was carried over from onboarding a Hotspot (to see if the Hotspot has an owner or not) and reused in this pairing flow. 

In any normal circumstance, this API check would not be a roadblock to pairing with a Hotspot, but during the outage, it was impossible to pair with a Hotspot, which exposed the API check. 

We don't think this api check is necessary and should be removed.

tyler-whitman:
We are checking for existing pending transactions when tapping the update hotspot item in hotspot settings which is taking a long time to complete now that the API is slow. 

I moved the check to happen when you actually try to submit the transaction instead. Allowing more time for a current pending to clear and also flows better with the UI.

closes #1093

### DIFF
--- a/src/features/hotspots/settings/updateHotspot/UpdateHotspotConfig.tsx
+++ b/src/features/hotspots/settings/updateHotspot/UpdateHotspotConfig.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from 'react'
+import React, { useCallback, useEffect, useMemo, useState } from 'react'
 import { Hotspot, Witness } from '@helium/http'
 import { useTranslation } from 'react-i18next'
 import { ActivityIndicator, Alert } from 'react-native'
@@ -34,6 +34,12 @@ import {
 import { calculateAssertLocFee } from '../../../../utils/fees'
 import { MakerAntenna } from '../../../../makers/antennaMakerTypes'
 import { isDataOnly } from '../../../../utils/hotspotUtils'
+import {
+  fetchTxnsHead,
+  HttpTransaction,
+} from '../../../../store/activity/activitySlice'
+import { isPendingTransaction } from '../../../wallet/root/useActivityItem'
+import { useAppDispatch } from '../../../../store/store'
 
 type Props = {
   onClose: () => void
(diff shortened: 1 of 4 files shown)


src/features/hotspots/settings/updateHotspot/UpdateHotspotConfig.tsx
```diff
@@ -183,8 +190,35 @@ const UpdateHotspotConfig = ({ onClose, onCloseSettings, hotspot }: Props) => {
...
+          pendingTxn.txn.gateway === hotspot?.address
+        )
+      })
+    } catch (e) {}
+    return false
+  }, [dispatch, hotspot?.address])
+
   const onSubmit = async () => {
     setLoading(true)
+    const hasExistingPendingTxn = await hasPendingTransaction()
```
matthewcarlreetz, line 216: should we do this on mount instead? That way if the API is slow it has more time to load.

src/utils/useHotspot.ts
```diff
@@ -129,6 +129,7 @@ const useHotspot = () => {
 
   const connectAndConfigHotspot = async (
     hotspotDevice: Device,
+    isDiagnostics?: boolean,
```
matthewcarlreetz, line 132: could we use an options object here? That way when viewing a function call the intent is more clear
`connectAndConfigHotspot(hotspot, { isDiagnostics: true} )`

matthewcarlreetz (approved):

src/features/hotspots/settings/updateHotspot/UpdateHotspotConfig.tsx
tyler-whitman: I feel like it would be weird to have it just pop up though. I think the flow is better so that it can wait until the last moment to check in case the other one clears.

src/utils/useHotspot.ts
tyler-whitman: Yeah, ill update it

src/features/hotspots/settings/updateHotspot/UpdateHotspotConfig.tsx
tyler-whitman: It has a nice loading animation in the button, so you at least know its loading

PR #1098, opened 2021-11-18
Add user agent to wallet client.

matthewcarlreetz:


src/utils/appDataClient.ts
```diff
@@ -18,15 +18,14 @@ import { fromNow } from './timeUtils'
 import * as Logger from './logger'
 
 const MAX = 100000
-const name =
-  Platform.OS === 'android' ? 'helium-wallet-android' : 'helium-wallet-ios'
+const userAgent = `helium-${Platform.OS}-js-client`
```
tyler-whitman, line 21: Do we want to add the app version?

matthewcarlreetz: yeah, that's a good idea

tyler-whitman (approved):

PR #1129, opened 2021-11-30
minor ui fixes

tyler-whitman:
- android splash screen set to cover mode
- fix transparent background on hotspot detail card


cokes518 (approved):
+1

PR #1157, opened 2021-12-08
integrate new discovery mode feature flags

tyler-whitman:


danielcolinjames (approved):
don't have context or a dev environment to run this locally but looks okay to me!

PR #1165, opened 2021-12-14
make sure scan view stops scanning after finding a qr

tyler-whitman:
Expo recommends setting the onBarCodeScanned to undefined after a scanning

https://github.com/expo/expo/issues/345

### DIFF
--- a/src/features/wallet/root/ActivityDetails/Payment.tsx
+++ b/src/features/wallet/root/ActivityDetails/Payment.tsx
@@ -42,16 +42,15 @@ const Payment = ({ item, address }: Props) => {
               index === payments.length - 1
             }
           />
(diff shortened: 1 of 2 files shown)


matthewcarlreetz:
Were you able to verify this works? I'm wondering if we should use the `useVisible` hook to conditionally render the camera.

tyler-whitman:
> Were you able to verify this works? I'm wondering if we should use the `useVisible` hook to conditionally render the camera.

Yeah, with this change there is no longer a lag when you scan multiple QRs in a row. Using state and setting the `onBarCodeScanned` to undefined after a scan is what expo recommends.

matthewcarlreetz (approved):

PR #1175, opened 2021-12-21
Added OPTION as a maker

RTM-Frank:
Added Option to the makers list for the CloudGate Helium Light-Hotspot.


cokes518:
This is a light hotspot maker and will likely need their own app by the time this hardware is supported on the network.

PR #1184, opened 2021-12-31
Show witnessed hotspots

### LINKED ISSUE #1182: Change "witnesses" to "witnessed" endpoint and average transmit scale of witnessed
our current app and explorer uses the /witnesses endpoint to determine how effective a Hotspot is at seeing other Hotspots. However, this metric is not a direct reflection of a Hotspot's ability to earn, in fact, the `/witnesses` endpoint just indicate what other Hotspots has seen the Hotspot in question, not the other way around (which is how rewards are determined).

To make this more clear and to better clarify the impact of transmit scale (i.e. your tx scale does not affect how much you earn), we will change to use the `/witnessed` endpoint, which lists the Hotspots that YOUR Hotspot has witnessed when they beacon. This is directly correlated with how much a Hotspot earns from witnessing rewards.

Furthermore, a Hotspot that can witness other Hotspots with a higher transmit scale means they get higher rewards. This is why we will show the **average** transmit scale of all Hotspots in the WitnessED list.

image

https://docs.helium.com/api/blockchain/hotspots#witnessed-for-a-hotspot

tyler-whitman:
This changes the witness list from a hotspots witnesses to instead show the hotspots that have witnessed it.

closes #1182


src/features/hotspots/details/HexBadge.tsx
```diff
@@ -54,13 +63,13 @@ const HexBadge = ({
...
+    if (!rewardScale) return 'grayText'
 
     return generateRewardScaleColor(rewardScale)
   }, [rewardScale])
 
   const scaleString = useMemo(() => {
-    if (!rewardScale) return ''
+    if (!rewardScale) return `0${decimalSeparator}00`
 
     if (rewardScale === 1) return `1${decimalSeparator}00`
```
matthewcarlreetz, line 74: Is `if (rewardScale === 1) return `1${decimalSeparator}00` still needed?

matthewcarlreetz (approved):

matthewcarlreetz:
I noticed one layout issue when font is scaled up

Simulator Screen Shot - iPhone 13 - 2022-01-04 at 11 01 17

src/features/hotspots/details/HexBadge.tsx
tyler-whitman: Yeah it's needed to format the text properly when the scale is 1.

matthewcarlreetz (approved):

PR #1190, opened 2022-01-06
Another attempt to fix hotspot settings black screen

tyler-whitman:
- remove hotspot settings spring animation
- remove blur box for hotspot settings
- simplify hotspot settings ui
- remove disable animations setting (was just for testing and didnt help)


allenan (approved):
let's try it

PR #1192, opened 2022-01-06
added People's Antenna - PPL

peoplesantenna:
* added manufacturer antenna People's Antenna - PPL

### DIFF
--- a/src/makers/index.ts
+++ b/src/makers/index.ts
@@ -69,7 +69,7 @@ export const AntennaModels = {
   ...syncrobit.antennas,
   ...kerlink.antennas,
   ...heltec.antennas,
-  ...peoples.antennas,
+  ...peoples,
   ...pisces.antennas,
   ...controllino.antennas,
   ...freedomfi.antennas,
--- a/src/makers/index.ts
+++ b/src/makers/index.ts
@@ -15,6 +15,7 @@ import sensecap from './sensecap'
 import syncrobit from './syncrobit'
 import kerlink from './kerlink'
 import heltec from './heltec'
+import peoples from './peoples/antennas'
 import pisces from './pisces'
 import controllino from './controllino'
 import freedomfi from './freedomfi'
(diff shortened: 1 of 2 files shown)


src/makers/peoples/antennas.ts
```diff
@@ -0,0 +1,8 @@
+import { MakerAntenna } from '../antennaMakerTypes'
+
+const peoples = {
+  name:'"People\'s Antenna" - PPL',
```
tyler-whitman, line 4: There is a lint error here. It should be `  name: '"People\'s Antenna" - PPL',`

src/makers/index.ts
```diff
@@ -68,6 +69,7 @@ export const AntennaModels = {
   ...syncrobit.antennas,
   ...kerlink.antennas,
   ...heltec.antennas,
+  ...peoples.antennas,
```
tyler-whitman, line 72: This should just be `...peoples`.

src/makers/peoples/antennas.ts
tyler-whitman: You may also want to remove the `"` so that this will be in the proper alphabetical order.

tyler-whitman (changes requested):

src/makers/peoples/antennas.ts
peoplesantenna: Fixed, thanks!

src/makers/index.ts
peoplesantenna: Fixed

peoplesantenna:
@tyler-whitman any idea when this will merge in? All the changes requested have been applied. Thank you!

cokes518:
Thanks for this PR but we will not be merging this. It was a mistake to add the initial third-party antenna (HNTenna) to the app in the first place. The only antennas that should be in the dropdown should be from HIP-19 approved manufacturers. As such, any other custom/third-party antenna will be removed from the app and future third-parties should use the custom antenna dropdown field. Users should be informed enough to input the dbi themselves.

https://github.com/helium/hotspot-app/pull/1234

PR #1201, opened 2022-01-10
ensure mock pending txns don't disappear

tyler-whitman:


src/store/activity/activitySlice.ts
```diff
@@ -310,7 +310,9 @@ const activitySlice = createSlice({
         state.txns[filter].hasInitialLoad = true
         if (filter === 'pending') {
           const pending = payload as HttpPendingTransaction[]
-          state.txns.pending.data = pending.filter(
+          const existingPending = state.txns.pending.data
+          const newPending = unionBy(pending, existingPending, 'hash')
```
matthewcarlreetz, line 314: Looks good, but I'm wondering why this was removed in the first place? Just want to avoid potentially bringing back a different bug.

matthewcarlreetz (approved):

src/store/activity/activitySlice.ts
tyler-whitman: Did it used to use union?

matthewcarlreetz: yes. It was removed in the cache PR 
https://github.com/helium/hotspot-app/blob/2e833240cc8af7c9a9ed744f997edeb7419552a0/src/store/activity/activitySlice.ts#L323

matthewcarlreetz: I don't recall any specific reason why it would have been removed.

tyler-whitman: It might have been a mistake in the refactor for cache then. I don't remember having a reason to remove it. I tested this a bunch and it seems to work well.

tyler-whitman: Maybe I was just simplifying it, but when the API is slow to get the pending transaction and we refresh quickly after submitting they dissappear.

PR #1211, opened 2022-01-14
ensure hotspot details are up to date when submitting assert txns

### LINKED ISSUE #1208: Location assert sets antenna gain and elevation to default
Users are reporting that location asserts are setting gain and elevation back to default values. Users should be able to update location for a $10 staking fee without affecting gain and elevation settings that were previously saved. The gain and elevation update should be a $0.55 fee without staking fees and should not change location.

tyler-whitman:
resolves #1208


allenan (approved):
looks good

PR #1216, opened 2022-01-20
add earnings map filter

tyler-whitman:
cloded #1210


src/components/Coverage.tsx
```diff
@@ -79,6 +80,7 @@ const Coverage = ({
   ownedHotspots,
   followedHotspots,
   showRewardScale,
+  showEarningsScale,
```
matthewcarlreetz, line 83: It feels like `showRewardScale` and `showEarningsScale` could be combined into one prop. Maybe a discriminated string union would be better.

src/features/map/MapFilterModal.tsx
```diff
@@ -27,11 +32,26 @@ type Props = {
...
+    (state: RootState) => state.rewards.networkHotspotEarnings.data,
+  )
   const [selectedFilter, setSelectedFilter] = useState(mapFilter)
 
+  useMount(() => {
+    dispatch(fetchNetworkHotspotEarnings())
+  })
+
+  const avg7dEarnings = useMemo(() => {
+    if (networkHotspotEarnings.length === 0) return
```
matthewcarlreetz, line 49: Probably not a big deal, but it's possible that the earnings job could fail and a day of earnings would be missing

matthewcarlreetz (approved):

src/features/map/MapFilterModal.tsx
tyler-whitman: This was the same logic as explorer, what would we want to do in the case it fails? It would always just take the last available 7, which should be a pretty close approximate in the case one day fails.

PR #1217, opened 2022-01-20
Add heliumrepublic antenna

loudadmin:


cokes518:
Thanks for this PR but we will not be merging this. It was a mistake to add the initial third-party antenna (HNTenna) to the app in the first place. The only antennas that should be in the dropdown should be from HIP-19 approved manufacturers. As such, any other custom/third-party antenna will be removed from the app and future third-parties should use the custom antenna dropdown field. Users should be informed enough to input the dbi themselves.

https://github.com/helium/hotspot-app/pull/1234

PR #1229, opened 2022-01-27
Add Dragino HP0D

dragino:
PR to add Dragino HP0D in APP


src/makers/dragino/index.ts
```diff
@@ -0,0 +1,9 @@
+import antennas from './antennas'
+import hotspots from './hotspots'
+
+export default {
+  antennas,
+  hotspots,
+  id: 27,
```
tyler-whitman, line 7: There is an extra space to remove here. You can find the lint errors by running `yarn lint`.
```
hotspot-app/src/makers/dragino/index.ts
  7:10  error  Delete `·`  prettier/prettier
```

tyler-whitman (changes requested):
Looks good, just a minor lint error to fix.

src/makers/dragino/index.ts
dragino: Just fix. Thanks!

tyler-whitman (approved):

PR #1248, opened 2022-02-18
Add code for QR onboard

MidasWireless:
Added code in hotspots.ts to support onboard through QR code.


cokes518:
Do you have your maker app ready to go? this seems better handled there than in the helium app (which will start deprecating Makers).

MidasWireless:
> Do you have your maker app ready to go? this seems better handled there than in the helium app (which will start deprecating Makers).

Our App will be available in 1 or 2 weeks, so it's better to have this PR merged for the transition period.

PR #1280, opened 2022-03-31
always use proxy for helium http client

tyler-whitman:


matthewcarlreetz:
@tyler-whitman  If I'm understanding the code correctly, we are removing the network selector and going back to `api.helium.io`?

matthewcarlreetz (approved):

tyler-whitman:
> @tyler-whitman If I'm understanding the code correctly, we are removing the network selector and going back to `api.helium.io`?

Correct, `helium-api.stakejoy.com` is no longer a valid url and now redirects to `api.helium.io`. Also see changes in https://github.com/helium/helium-js/pull/288

PR #1295, opened 2022-04-20
Fix the bug that can't open the currency selector

llwslc:
If the `currencyType` is not in the `SUPPORTED_CURRENCIES` list, `Currency` selector will not be opened

### DIFF
--- a/src/features/moreTab/more/MoreScreen.tsx
+++ b/src/features/moreTab/more/MoreScreen.tsx
@@ -40,6 +40,7 @@ import AppInfoItem from './AppInfoItem'
 import DeployModeModal from './DeployModeModal'
 import { useLanguageContext } from '../../../providers/LanguageProvider'
 import { EXPLORER_BASE_URL } from '../../../utils/config'
+import { currencyType as defaultCurrencyType } from '../../../utils/i18n'
 import { SUPPORTED_LANGUAGUES } from '../../../utils/i18n/i18nTypes'
 import Articles from '../../../constants/articles'
 import useAlert from '../../../utils/useAlert'
@@ -56,6 +57,9 @@ const MoreScreen = () => {
   const showHiddenHotspots = useSelector(
     (state: RootState) => state.account.settings.showHiddenHotspots,
   )
+  const currencyType =
+    useSelector((state: RootState) => state.account.settings.currencyType) ||
+    defaultCurrencyType
   const account = useSelector((state: RootState) => state.account, isEqual)
   const fleetModeLowerLimit = useSelector(
     (state: RootState) => state.features.fleetModeLowerLimit,
@@ -288,6 +292,11 @@ const MoreScreen = () => {
         disabled: app.isDeployModeEnabled,
       },
     ]
+
+    if (!SUPPORTED_CURRENCIES[currencyType]) {
+      SUPPORTED_CURRENCIES[currencyType] = currencyType
+    }
+    
     return [
       {
         title: t('more.sections.security.title'),
@@ -331,7 +340,7 @@ const MoreScreen = () => {
           },
           {
             title: t('more.sections.app.currency'),
-            value: account.settings.currencyType,
+            value: currencyType,
             select: {
               items: Object.keys(SUPPORTED_CURRENCIES).map((p) => {
                 return {
@@ -387,7 +396,7 @@ const MoreScreen = () => {
     handleRevealWords,
     language,
     handleLanguageChange,
-    account.settings.currencyType,
+    currencyType,
     account.settings.convertHntToCurrency,
     account.settings.isFleetModeEnabled,
     handleCurrencyTypeChange,


src/features/moreTab/more/MoreScreen.tsx
```diff
@@ -40,6 +40,7 @@ import AppInfoItem from './AppInfoItem'
 import DeployModeModal from './DeployModeModal'
 import { useLanguageContext } from '../../../providers/LanguageProvider'
 import { EXPLORER_BASE_URL } from '../../../utils/config'
+import { currencyType as defaultCurrencyType } from '../../../utils/i18n'
```
tyler-whitman, line 43: This is a string array, where currency type expects a string. Lets update the i18n file to have a defaultCurrencyType of USD. On line 17 of `../../../utils/i18n` add:
```
export const defaultCurrencyType = 'USD'
export const [currencyType] = RNLocalize.getCurrencies() || [
  defaultCurrencyType,
]
```

You can then change this import to:
```
import { defaultCurrencyType } from '../../../utils/i18n'
```

src/features/moreTab/more/MoreScreen.tsx
```diff
@@ -56,6 +57,9 @@ const MoreScreen = () => {
   const showHiddenHotspots = useSelector(
     (state: RootState) => state.account.settings.showHiddenHotspots,
   )
+  const currencyType =
```
tyler-whitman, line 60: We can then remove this, and just use a check on line 343

src/features/moreTab/more/MoreScreen.tsx
```diff
@@ -331,7 +340,7 @@ const MoreScreen = () => {
           },
           {
             title: t('more.sections.app.currency'),
-            value: account.settings.currencyType,
+            value: currencyType,
```
tyler-whitman, line 343: We can just make this:

```
account.settings.currencyType || defaultCurrencyType
```

src/features/moreTab/more/MoreScreen.tsx
```diff
@@ -288,6 +292,11 @@ const MoreScreen = () => {
         disabled: app.isDeployModeEnabled,
       },
     ]
+
+    if (!SUPPORTED_CURRENCIES[currencyType]) {
```
tyler-whitman, line 296: I dont believe this should be needed if we are setting `defaultCurrencyType` to USD

tyler-whitman (changes requested):
Thanks for the fix, have a few changes to address.

src/features/moreTab/more/MoreScreen.tsx
tyler-whitman: On second thought, this should actually already be defaulting to USD in the `accountSlice.ts` if `RNLocalize.getCurrencies()` returned undefined. 

In order to encounter this bug, what is being returned by `RNLocalize.getCurrencies()`. I think the actual fix will be to check if the currency returned by `RNLocalize.getCurrencies()` is one of the supported currencies and if not we need to default to USD. We can only support currencies that are listed on coingecko via its API `https://api.coingecko.com/api/v3/coins/helium?localization=false&tickers=false&community_data=false&developer_data=false`

tyler-whitman, line 43: This should already be defaulting to USD in the accountSlice.ts if RNLocalize.getCurrencies() returned undefined.

In order to encounter this bug, what is being returned by RNLocalize.getCurrencies(). I think the actual fix will be to check if the currency returned by RNLocalize.getCurrencies() is one of the supported currencies and if not we need to default to USD. We can only support currencies that are listed on coingecko via its API https://api.coingecko.com/api/v3/coins/helium?localization=false&tickers=false&community_data=false&developer_data=false

llwslc: I found this problem because 'RNLocalize.getCurrencies()' returned 'CNY'. We can set all unsupported currencies to USD, but if coingecko adds a new fiat currency, and we can only support it by updating the app, that's a bit of a pain.

src/features/moreTab/more/MoreScreen.tsx
llwslc: I found out that 'CNY' failed.

src/features/moreTab/more/MoreScreen.tsx
llwslc: I just checked, although we support Japanese and Korean, we also do not support `JPY` and `KRW`. Coingecko supports both. Also users in both places can't select USD, because they can't open the currency selector.

tyler-whitman: So we are missing a check to ensure that the users currency is not in the supported currencies list, and we can update the supported currencies to include a few more if they are listed in the coingecko api

llwslc: Yes, I look forward to a more reasonable solution from you. I'm closing this merge for now. Thanks a lot!

PR #1305, opened 2022-05-12
light hotspot updates

### LINKED ISSUE #1304: Light Hotspot changes for App
1. After Light hotspots, there is no concept of relay. we can remove that copy all together.
2. Remove NAT type from diagnostics
3. If we have copy related to sync/syncing, we can update that to say "Connected". If a Hotspot is offline, it can continue to say "Offline".
4. Double check block sync check for transfer Hotspots. If there is one, remove it. we should rely on poc activity check.

tyler-whitman:
closes #1304


src/features/hotspots/settings/discovery/DiscoveryModeRoot.tsx
```diff
@@ -35,7 +32,6 @@ const DiscoveryModeRoot = ({ onClose, hotspot }: Props) => {
   const [errorShownForRequestId, setErrorShownForRequestId] = useState<
```
matthewcarlreetz, line 32: Disco Mode 😢

matthewcarlreetz (approved):
LGTM

PR #1334, opened 2022-07-18
add Aitek Maker

aitekinc:
Add Aitek maker


cokes518:
Hi, this PR is not needed as your customers should be using your maker app with deeplink enabled.

aitekinc:
Hi Coco, we do recommend the customers to use the maker app , just want to list the rest of the makers on the helium app as well, for customers' convenience.

Issue #1341, opened 2022-08-05
Allow QR code scanning of payment_v2 transactions


rawrmaan:
The app should allow scanning a QR code to initiate a `transfer_hotspot_v2` transaction. Importantly, the client-side check for recent hotspot activity (normally required to transfer a hotspot) should be bypassed so advanced users can make their own determination of whether a hotspot is ready to transfer.

This is the screen that pushes the transfer UI (see `navigateToTransfer` function): https://github.com/helium/hotspot-app/blob/main/src/features/hotspots/transfers/HotspotTransfer.tsx

And the main transfer UI is in `SendView`. In order to bypass the activity check, scanned transfers could follow the same code path as data-only hotspots, seen here: `https://github.com/helium/hotspot-app/blob/afbe3342b4b663787ae379e143a145704bd32067/src/features/wallet/send/SendView.tsx#L163`

PR #1342, opened 2022-08-07
[#1341] feat: Allow QR code scanning of "transfer_hotspot_v2" transactions

MichaelBrew:
Issue: https://github.com/helium/hotspot-app/issues/1341

This PR adds the ability to scan a QR code allowing users to pre-fill the "Transfer Hotspot" form (while also skipping the "has valid activity" check if directed as such in the code itself). The expected payload for the newly supported QR code is:
```
{
  type: "transfer"
  newOwnerAddress: string
  hotspotAddress?: string
  skipActivityCheck?: boolean
  isSeller?: boolean
}
```

This PR also updates `src/providers/AppLinkProvider.tsx` to determine a specific "ScanDataType" for scanned data, then handles each individually
```typescript
// ScanDataType describes both the scanned data format as well as intended use.
// For example, multiple types are intended to pre-fill the payment "send" form but can provide
// varying levels of information (like single or multiple recipients, optional memo fields, etc).
enum ScanDataType {
  // Deeplink to somewhere in the app
  DEEPLINK,
  // Hotspot address update
  LOCATION_UPDATE,
  // DC burn
  DC_BURN,
  // Hotspot transfer
  TRANSFER,
  // Hotspot transfer with only address provided
  TRANSFER_ADDRESS_ONLY,
  // Payment
  PAYMENT,
  // Payment with only address provided
  PAYMENT_ADDRESS_ONLY,
  // Payment to multiple recipients
  PAYMENT_MULTI,
  // Payment to multiple recipients and optional memos
  PAYMENT_MULTI_MEMO,
}
```
image

### DIFF
--- a/src/providers/appLinkTypes.ts
+++ b/src/providers/appLinkTypes.ts
@@ -23,7 +23,15 @@ export type AppLink = {
   address: string
   amount?: string | number
   memo?: string
-  [key: string]: string | number | undefined
+  [key: string]: string | number | boolean | undefined
+}
+
+export type AppLinkTransfer = {
+  type: 'transfer'
+  newOwnerAddress: string
+  hotspotAddress?: string
+  skipActivityCheck?: boolean
+  isSeller?: boolean
 }
 
 export type AppLinkPayment = {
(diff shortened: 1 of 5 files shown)


src/providers/appLinkTypes.ts
```diff
@@ -23,7 +23,15 @@ export type AppLink = {
   address: string
   amount?: string | number
   memo?: string
-  [key: string]: string | number | undefined
```
kenstuhh, line 26: [key: string]: string | number | boolean | undefined

kenstuhh (approved):

kenstuhh (approved):

PR #1367, opened 2022-12-09
App migration changes

tyler-whitman:
- Show sentinel every 24 hours on app launch
- Preserve 12 words when migrating to wallet app
- Adds an alert before transferring or updating hotspots letting you know about the Wallet App
- Updates more of the maker links


matthewcarlreetz (approved):
Looks good to me!

PR #1373, opened 2022-12-16
react native 70

tyler-whitman:


Perronef5 (approved):
LGTM. Only thing i'm wondering is if the `.ruby-version` is required?

tyler-whitman:
> LGTM. Only thing i'm wondering is if the `.ruby-version` is required?

That’s something that react native adds. I see it in the upgrade change list so let’s leave it be haha

